PR #3, opened 2022-02-15
Fix tag workflow suffix

pshapiro4broad:
~The tag operation adds a `-` to the suffix so the GitHub action input doesn't need to have it.~

The `SNAPSHOT` suffix isn't necessary after all, see comment below.


fboulnois (approved):

pshapiro4broad:
I'm now wondering if `SNAPSHOT` is correct at all. I think the intent is that `-SNAPSHOT` is used for on-branch container releases. In that case, a released image wouldn't have any suffix at all. What do you think @fboulnois ?

samanehsan (approved):
Lgtm!

PR #4, opened 2022-02-15
DC-116: Bump MINOR instead of PATCH

fboulnois:
Per the Semantic Versioning standard, the MINOR version should be bumped when you add functionality in a backwards compatible manner.

### DIFF
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -11,7 +11,7 @@ jobs:
           token: ${{ secrets.BROADBOT_TOKEN }} # this allows the push to succeed later
       - name: Bump the tag to a new version
         # https://github.com/DataBiosphere/github-actions/tree/master/actions/bumper
-        uses: databiosphere/github-actions/actions/bumper@bumper-0.0.3
+        uses: databiosphere/github-actions/actions/bumper@bumper-0.0.4
         id: tag
         env:
           GITHUB_TOKEN: ${{ secrets.BROADBOT_TOKEN }}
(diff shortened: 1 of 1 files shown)


samanehsan (approved):

pshapiro4broad:
My thinking was that we'd bump patch by default and then bump minor before a release, possibly by manually running the tag workflow. That way we could keep the minor version rev number relatively low. 

On the other hand, we could in theory release at any time so maybe always bumping minor instead would be better.

fboulnois:
> My thinking was that we'd bump patch by default and then bump minor before a release, possibly by manually running the tag workflow. That way we could keep the minor version rev number relatively low.
> 
> On the other hand, we could in theory release at any time so maybe always bumping minor instead would be better.

A new change in the checklist compared to when we did it for Data Repo is that new services MUST follow SemVer.

According to SemVer, 
> Minor version Y (x.Y.z | x > 0) MUST be incremented if new, backwards compatible functionality is introduced to the public API. It MUST be incremented if any public API functionality is marked as deprecated. It MAY be incremented if substantial new functionality or improvements are introduced within the private code. It MAY include patch level changes. Patch version MUST be reset to 0 when minor version is incremented.
> Patch version Z (x.y.Z | x > 0) MUST be incremented if **only** backwards compatible bug fixes are introduced. A bug fix is defined as an internal change that fixes incorrect behavior.

Therefore, since we're mostly going to be adding new features, I think it makes more sense to bump the MINOR version by default, and use the override with a PATCH version if we're committing a backwards compatible bugfix.

pshapiro4broad (approved):

PR #7, opened 2022-02-18
DC-236: Add Trivy vulnerability scanning

fboulnois:


.github/workflows/trivy.yml
```diff
@@ -0,0 +1,30 @@
...
+      - uses: actions/checkout@v2
+      - name: Set up JDK
+        uses: actions/setup-java@v2
+        with:
+          java-version: '17'
+          distribution: 'temurin'
+
+      - name: Construct docker image name and tag
+        id: image-name
+        run: echo ::set-output name=name::${{ github.event.repository.name }}:$(git rev-parse --short HEAD)
```
pshapiro4broad, line 17: Generating a tag seems unnecessary for this workflow since the image isn't pushed. You could use a fixed tag (or no tag at all) for the purposes of this scan.

pshapiro4broad (approved):
Could this be done as part of the Build and Test workflow? That workflow is run on PRs and is already building the jib container.

.github/workflows/trivy.yml
fboulnois: The image isn't pushed, but both the build step and scanning step need the same image name, and this is less hacky then reading the output of the build log.

samanehsan (approved):

PR #9, opened 2022-02-18
DC-229: Fix recursive workflow

fboulnois:
When this action runs on push to main and succeeds, it will launch the tag bumping action. The tag bumping action will push a new version to main, causing this workflow to run again recursively.


pshapiro4broad (approved):
This seems fine to me, but I wonder why this fix is needed. I copied these workflows from another repo where they seem to be working. Maybe they have some other config setting that we don't? Or maybe they weren't working correctly there either.

PR #11, opened 2022-02-22
DC-229: Fix Slack webhook URL

fboulnois:


samanehsan (approved):

pshapiro4broad (approved):
Could you also add a custom user name for both slack integrations? Right now it will post as `jade-spam incoming-webhook`. The attribute is `username`, one possible setting would be:
```
          username: 'GitHub Workflow'
```

PR #13, opened 2022-02-23
DC-141: Run action nightly

fboulnois:


.github/workflows/nightly-perf-tests.yml
```diff
@@ -0,0 +1,145 @@
...
+          if [[ $resultStatus == 0 ]]; then
+            echo "Server started successfully"
+          else
+            echo "Server did not start successfully. Dumping log file"
+            cat bootrun.log
+            exit 1
+          fi
+
+      - name: Run the integration test suite
+        run: ./gradlew --build-cache :integration:runTest --args="suites/FullIntegration.json /tmp/foo"
```
pshapiro4broad, line 126: The test suite is different for perf
```suggestion
        run: ./gradlew --build-cache :integration:runTest --args="suites/FullPerf.json /tmp/foo"
```

.github/workflows/nightly-perf-tests.yml
```diff
@@ -0,0 +1,145 @@
...
+      - name: Gradle cache
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          key: v1-${{ runner.os }}-gradle-${{ github.ref }}-${{ github.sha }}
+          restore-keys: v1-${{ runner.os }}-gradle-${{ github.ref }}
+
+      - name: Gradle build service
```
pshapiro4broad, line 27: I think the nightly perf test should run against the perf environment, which (should) already have a catalog service deployed to it. If so, some of this setup won't be necessary; we don't need to build or run postgres, for example.

pshapiro4broad (changes requested):
I think perf tests are set up differently than the integration tests. A possible example of this workflow is https://github.com/DataBiosphere/terra-resource-buffer/blob/master/.github/workflows/test-runner-nightly-perf.yml

.github/workflows/nightly-perf-tests.yml
pshapiro4broad: The catalog config for perf is using the perf server (not localhost):

https://github.com/DataBiosphere/terra-data-catalog/blob/main/integration/src/main/resources/servers/catalog-perf.json#L6

(ignore the use of `policyManagerUri`, this will be fixed in a future PR)

.github/workflows/nightly-perf-tests.yml
fboulnois: Yes, I agree. I thought we would run both integration and perf tests? The way I wrote it so far would only cover the integration tests.

pshapiro4broad: We haven't yet talked about what's in a perf test vs what's in an integration test. One possible approach is to have perf include everything integration does, but with more simultaneous operations.

If we do have a nightly test that runs integration tests, maybe that should use the dev deployment instead of running a local deployment? This can be done using an env var to override the server config setting for the integration tests.

samanehsan (approved):
Lgtm :octocat:

.github/workflows/nightly-perf-tests.yml
```diff
@@ -0,0 +1,111 @@
...
+        run: ./gradlew --build-cache :service:build -x test
+
+  jib:
```
pshapiro4broad, line 31: `jib` doesn't appear to be used by any downstream steps, except `notify-slack`. I don't think it's needed for this workflow.

Once `jib` is removed, I think `build` can be removed too. The `integration` subproject doesn't depend on `service` so building `service` shouldn't have any effect on running the perf tests.

fboulnois: `jib` is needed for Trivy to run against the generated image

pshapiro4broad: OK. Although now we're running trivy on PRs, after PR merge, and nightly. Would it be possible / worthwhile to put the trivy scan in its own workflow, so the workflow code isn't duplicated in the nightly and and PR workflows?

pshapiro4broad (approved):

.github/workflows/nightly-perf-tests.yml
fboulnois: A separate Trivy action would have to build the code and containerize it, so the build section would be duplicated across all the actions. As a result I'm not sure if that's worth separating it out into another action, especially since we leverage the build-cache for the actions, whereas a separate Trivy action would build from scratch.

PR #15, opened 2022-02-24
[DC-234] Add status and version unit tests

samanehsan:


service/src/test/java/bio/terra/catalog/api/StatusTest.java
```diff
@@ -0,0 +1,43 @@
...
+import org.springframework.boot.test.autoconfigure.web.servlet.WebMvcTest;
+import org.springframework.boot.test.mock.mockito.MockBean;
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.context.junit.jupiter.SpringExtension;
+import org.springframework.test.web.servlet.MockMvc;
+
+@ExtendWith(SpringExtension.class)
+@ContextConfiguration(classes = PublicApiController.class)
+@WebMvcTest
+public class StatusTest {
```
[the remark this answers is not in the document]
fboulnois: That seems like a good idea, especially since down the line we might add more stuff in PublicApiController

service/src/test/java/bio/terra/catalog/api/VersionTest.java
```diff
@@ -0,0 +1,56 @@
...
+    when(mockConfig.getGithub()).thenReturn("https://github.com/");
+    when(mockConfig.getBuild()).thenReturn("build");
+    return mockConfig;
+  }
+}
+
+@ExtendWith(SpringExtension.class)
+@ContextConfiguration(classes = {PublicApiController.class, TestVersionConfig.class})
+@WebMvcTest
+public class VersionTest {
```
pshapiro4broad, line 40: A minor point; JUnit 5 doesn't require test classes or methods to be `public` like 4 did. So the junit5 recommendation is to use default access for classes and methods. https://junit.org/junit5/docs/current/user-guide/#writing-tests-classes-and-methods

.github/workflows/build-and-test.yml
```diff
@@ -110,7 +110,7 @@ jobs:
       - name: Codecov
         uses: codecov/codecov-action@v2.0.2
         with:
-          files: ./service/build/reports/jacoco/test/jacocoTestReport.xml
+          files: ./service/build/reports/jacoco/test/html/index.html
```
samanehsan, line 113: `jacoco` won't create this file until the unit tests run so the codecov job fails. I'll open a separate ticket for that!

pshapiro4broad (approved):

fboulnois (approved):
lgtm

PR #17, opened 2022-02-28
DC-251: Add source clear and sonarqube scanners to build

pshapiro4broad:
This adds source clear and sonarqube scans to our PR tests. Source clear checks for vulnerabilities in .jars that we depend on, and sonarqube runs a number of static code analysis checks on our Java code.

~Note that builds will fail until the GitHub secrets are published once https://github.com/broadinstitute/terraform-ap-deployments/pull/551 is merged.~

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -120,14 +120,7 @@ jobs:
         with:
           java-version: '17'
           distribution: 'temurin'
-
-      - name: Gradle cache
-        uses: actions/cache@v2
-        with:
-          path: |
-            ~/.gradle/caches
-            ~/.gradle/wrapper
-          key: v1-${{ runner.os }}-gradle-${{ github.ref }}-${{ github.sha }}
+          cache: 'gradle'
 
       - name: SonarQube scan
         run: ./gradlew --build-cache -Dsonar.login="${{ secrets.SONAR_TOKEN }}" sonarqube
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -151,7 +151,7 @@ jobs:
           key: v1-${{ runner.os }}-gradle-${{ github.ref }}-${{ github.sha }}
 
       - name: SonarQube scan
-        run: ./gradlew -Dsonar.login="${{ secrets.SONAR_TOKEN }}" --build-cache sonarqube
+        run: ./gradlew --build-cache -Dsonar.login="${{ secrets.SONAR_TOKEN }}" sonarqube
 
   integration-tests:
     needs: [ build ]
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -112,6 +112,54 @@ jobs:
         with:
           files: ./service/build/reports/jacoco/test/jacocoTestReport.xml
 
+  source-clear:
+    needs: [ build ]
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up JDK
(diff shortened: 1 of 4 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -112,6 +112,54 @@ jobs:
...
+      - name: Gradle cache
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          key: v1-${{ runner.os }}-gradle-${{ github.ref }}-${{ github.sha }}
+
+      - name: SonarQube scan
+        run: ./gradlew -Dsonar.login="${{ secrets.SONAR_TOKEN }}" --build-cache sonarqube
```
samanehsan, line 161: I'm getting confused about the global vs. service build.gradle files. If the sonarqube task is defined in the service, why isn't this :service:sonarqube? 🤔

buildSrc/build.gradle
```diff
@@ -12,10 +12,12 @@ repositories {
 dependencies {
     implementation 'com.diffplug.spotless:spotless-plugin-gradle:6.3.0'
     implementation 'com.google.cloud.tools.jib:com.google.cloud.tools.jib.gradle.plugin:3.2.0'
+//    implementation 'com.srcclr.gradle:com.srcclr.gradle.gradle.plugin:3.1.10'
```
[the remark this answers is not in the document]
pshapiro4broad: Yes, forgot to delete that.

.github/workflows/build-and-test.yml
pshapiro4broad: Either would work in this case. The task is only run in the subprojects that have it present, so using a global task means that if we add this scan to other projects, we don't have to update the Gradle command. This seems like a nice feature to me, but I could make it more specific instead if you think it's confusing.

samanehsan: Thanks for clarifying! I think we can leave it as-is so we don't have to update the command later

.github/workflows/build-and-test.yml
```diff
@@ -112,6 +112,54 @@ jobs:
...
+          java-version: '17'
+          distribution: 'temurin'
+
+      - name: Gradle cache
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          key: v1-${{ runner.os }}-gradle-${{ github.ref }}-${{ github.sha }}
```
fboulnois, line 133:
```suggestion
          distribution: 'temurin'
          cache: 'gradle'
```

.github/workflows/build-and-test.yml
```diff
@@ -112,6 +112,54 @@ jobs:
...
```
fboulnois, line 158:
```suggestion
          distribution: 'temurin'
          cache: 'gradle'
```

.github/workflows/build-and-test.yml
fboulnois, line 161:
```suggestion
        run: ./gradlew --build-cache -Dsonar.login="${{ secrets.SONAR_TOKEN }}" sonarqube
```

nit: lead with `--build-cache` for consistency with other items?

fboulnois (changes requested):
A few suggestions:

.github/workflows/build-and-test.yml
fboulnois: Also, does the code need to be containerized beforehand for any of these tools to work correctly?

fboulnois (approved):
lgtm

samanehsan (approved):

PR #20, opened 2022-03-01
DC-233: update to use testrunner field

pshapiro4broad:
Now that testrunner has a server config for catalog, use it instead of policy manager.


samanehsan (approved):
Lgtm! :octocat:

fboulnois (approved):
lgtm

PR #21, opened 2022-03-02
DC-240: First pass at README and getting started docs

fboulnois:
Adds sections in the README for building and testing code, and adds a getting started document for totally new team members.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
 # Terra Data Catalog
 
 [![Build and Test](https://github.com/DataBiosphere/terra-data-catalog/actions/workflows/build-and-test.yml/badge.svg?branch=main)](https://github.com/DataBiosphere/terra-data-catalog/actions/workflows/build-and-test.yml)
+[![Nightly Tests](https://github.com/DataBiosphere/terra-data-catalog/actions/workflows/nightly-tests.yml/badge.svg)](https://github.com/DataBiosphere/terra-data-catalog/actions/workflows/nightly-tests.yml)
+[![Publish and deploy](https://github.com/DataBiosphere/terra-data-catalog/actions/workflows/publish.yml/badge.svg)](https://github.com/DataBiosphere/terra-data-catalog/actions/workflows/publish.yml)
 
 The mission of the Terra Data Catalog is to make research data accessible and
 searchable to accelerate biomedical discoveries.
@@ -11,11 +13,11 @@ searchable to accelerate biomedical discoveries.
 first.
 
 Ensure you have Java 17 and that it is the default. To check this while in the
-`terra-data-catalog` directory, type `./gradlew --version`.
+`terra-data-catalog` directory, type `java --version`.
 
 Then, to build the code, run:
 
-```
+```sh
 ./gradlew build
 ```
 
@@ -24,13 +26,14 @@ Then, to build the code, run:
 For tests, ensure you have a local Postgres instance running. While in the
 `terra-data-catalog` directory, initialize the database as follows:
 
-```
+```sh
 psql -f common/postgres-init.sql
 ```
 
 After the database is initialized, then you may run integration tests as follows:
 
-```
-render-configs.sh
-./gradlew :integration:runTest --scan --args="suites/FullIntegration.json /tmp/test"
+```sh
+./gradlew bootRun & # start up a local instance of the data catalog service
+render-configs.sh # render service account credentials needed for tests
+./gradlew :integration:runTest --args="suites/FullIntegration.json /tmp/test"
 ```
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -4,3 +4,33 @@
...
+
+```
+psql -f common/postgres-init.sql
+```
+
+After the database is initialized, then you may run integration tests as follows:
+
+```
+render-configs.sh
+./gradlew :integration:runTest --scan --args="suites/FullIntegration.json /tmp/test"
```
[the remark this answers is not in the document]
pshapiro4broad: Also note `--scan` isn't necessary to run the integration tests.

pshapiro4broad (approved):
Looks good, just a few minor comments

docs/getting-started.md
```diff
@@ -0,0 +1,184 @@
...
+# configure vault
+export VAULT_ADDR=https://clotho.broadinstitute.org:8200
+
+# launch docker desktop - this installs docker in /usr/local/bin
+open -a docker
+
+# configure google-cloud-sdk
+gcloud auth login
+gcloud auth application-default login
+gcloud auth configure-docker
```
pshapiro4broad, line 122: I think these are only needed if you're working on debugging k8s or helm. We don't use BQ and our app itself doesn't auth with google directly (I think). I guess they'd be useful if you want to proxy to our databases for debugging an issue in dev or prod.

fboulnois: I think this is useful for debugging later on, so I'll keep these in for now.

README.md
```diff
@@ -24,13 +26,14 @@ Then, to build the code, run:
...
 psql -f common/postgres-init.sql
 ```
 
 After the database is initialized, then you may run integration tests as follows:
 
-```
-render-configs.sh
-./gradlew :integration:runTest --scan --args="suites/FullIntegration.json /tmp/test"
+```sh
+./gradlew bootRun & # start up a local instance of the data catalog service
```
lu-c, line 36: What does the `&` do in this command? Is it necessary?

README.md
```diff
@@ -24,13 +26,14 @@ Then, to build the code, run:
...
+render-configs.sh # render service account credentials needed for tests
+./gradlew :integration:runTest --args="suites/FullIntegration.json /tmp/test"
 ```
```
lu-c, line 39: Can we also add this to the readme?

```suggestion
```# Run Performance Tests:
```./render-configs.sh perf
```./gradlew :integration:runTest --args="suites/FullPerf.json /tmp/test"
```

README.md
fboulnois: Yes, it's necessary so that there's an instance that runs locally so that the tests can run against it.

pshapiro4broad: I actually wouldn't recommend running it this way, since `bootRun` is very noisy. I would recommend running it in a separate terminal so you can see its output separate from the test run output.

docs/getting-started.md
```diff
@@ -0,0 +1,186 @@
...
+
+> During this process, you will need your GitHub username, password, and
+personal access token for multiple steps, so make sure to have those handy. If
+you don't have those yet, see the section below, otherwise you can skip to
+[Request Required Access](#2-request-required-access).
+
+## 1. Create a GitHub account
+
+GitHub is where the Broad stores our code and projects. Sign up to this service
+with your **personal** email: https://github.com/join
```
pshapiro4broad, line 15: I never noticed this before. You can use your personal email, or you could use your broad email. If you use your broad account, then your work GitHub account will be separate from your personal one, which may be what a user wants to do. The important thing is that the GitHub account is linked to your broad account using https://github.broadinstitute.org

docs/getting-started.md
```diff
@@ -0,0 +1,186 @@
...
+5. To avoid being overwhelmed with notifications, [add your Broad email address](https://github.com/settings/emails),
+[route the notifications](https://github.com/settings/notifications) to that
+email, and [unfollow projects](https://github.com/watching) that are not
+relevant to your team.
+
+## 4. Create Terra Accounts
+
+The Data Catalog and [Terra](https://terra.bio/) use [Sam](https://github.com/broadinstitute/sam)
+to abtract identity and access management. To gain access to these services,
+first create a non-Broad email address through Gmail. This email address will
```
pshapiro4broad, line 58: "a non-Broad email address" -- maybe provide a suggestion here so people's accounts follow a convention? I created mine when I was helping out the monster team so mine is `monster.pshapiro.dev@gmail.com`

lu-c (approved):
LGTM!

docs/getting-started.md
fboulnois: I'm not sure we should suggest anything specific here. I started including this in the document, but we should support any email in general for Terra accounts. I'm concerned that any convention we set will become a standard that something will depend on.

PR #22, opened 2022-03-03
[DC-260] Add dataset table and DAO layer

samanehsan:
- Define the schema in liquibase (note: the `pgcrypto` extension is required to use the uuid_function)
- Add a DAO class with basic CRUD operations that directly uses jdbc. Currently the create and update methods take in the `Dataset` record model, but we'll want to change this to the request model once we have the equivalent API endpoints implemented.


service/src/main/java/bio/terra/catalog/service/dataset/Dataset.java
```diff
@@ -0,0 +1,57 @@
+package bio.terra.catalog.service.dataset;
+
+import java.time.Instant;
+import java.util.UUID;
+
+public class Dataset {
```
pshapiro4broad, line 6: Can you use a java `record` for this? It would look something like
```
public record Dataset(UUID id, String datasetId, String storageSystem, String metadata, Instant createdDate) { }
```

service/src/main/java/bio/terra/catalog/service/dataset/DatasetDao.java
```diff
@@ -0,0 +1,80 @@
...
+import org.springframework.transaction.annotation.Propagation;
+import org.springframework.transaction.annotation.Transactional;
+
+@Repository
+public class DatasetDao {
+
+  private final NamedParameterJdbcTemplate jdbcTemplate;
+
+  @Autowired
+  public DatasetDao(NamedParameterJdbcTemplate jdbcTemplate) throws SQLException {
```
pshapiro4broad, line 25: Is `throws SQLException ` necessary?

service/src/main/java/bio/terra/catalog/service/dataset/Dataset.java
```diff
@@ -0,0 +1,57 @@
...
+  private UUID id;
+  private String datasetId;
+  private String storageSystem;
```
pshapiro4broad, line 9: At some layer this will be an enum, I think. If we can make it an enum at this layer, it will probably be better, to avoid having to convert it in a caller method.

Going down to the database, we could make this a constrained field in SQL and only allow valid enum values in the database. I don't feel strongly about that level of validation, but it's possible to do.

service/src/test/java/bio/terra/catalog/service/dataset/DatasetDaoTest.java
```diff
@@ -0,0 +1,68 @@
...
+        new Dataset()
+            .datasetId(datasetId)
+            .storageSystem(String.valueOf(storageSystem.TERRA_WORKSPACE));
+    try {
+      datasetDao.create(dataset);
+      datasetDao.create(duplicateDatasetId);
+      int datasetCount = datasetDao.enumerate().size();
+      assertEquals(datasetCount, 2);
+    } finally {
+      datasetDao.enumerate().forEach(d -> datasetDao.delete(d.getId()));
```
pshapiro4broad, line 48: If the IDs are random, do we need to delete the data created? The unit tests will only be run against local databases (that is, never against dev or other persistent environments). I think it would be better to not have each test clean up after itself, in order to keep the test code simple.

samanehsan: We could move the delete to a separate cleanup method to simplify the tests, but I don't feel strongly either way

service/src/main/java/bio/terra/catalog/service/dataset/Dataset.java
samanehsan: That looks so much better!

service/src/main/java/bio/terra/catalog/service/dataset/exception/DatasetNotFoundException.java
```diff
@@ -0,0 +1,9 @@
+package bio.terra.catalog.service.dataset.exception;
+
+import bio.terra.common.exception.NotFoundException;
+
+public class DatasetNotFoundException extends NotFoundException {
```
samanehsan, line 5: The sonarcloud check fails because these exceptions have 6 parent classes! I'm wondering if I should just use the parent class instead @pshapiro4broad?

samanehsan:
The sonarcloud coverage is 0% because the unit tests aren't running in the action yet

service/src/main/java/bio/terra/catalog/service/dataset/exception/DatasetNotFoundException.java
pshapiro4broad: I think this can be fixed by customizing the scan to treat `ErrorReportException` as the root of the inheritance tree. I'll look at doing that now.

service/src/main/resources/db/changesets/20220301_datasettable.yaml
```diff
@@ -0,0 +1,40 @@
...
+        - createTable:
+            tableName: dataset
+            columns:
+              - column:
+                  name: id
+                  type: ${uuid_type}
+                  defaultValueComputed: ${uuid_function}
+                  constraints:
+                    primaryKey: true
+                    nullable: false
```
fboulnois, line 15: All the operation are done through this `uuid` as a primary key. Right now it is possible to have multiple duplicate `dataset_id` and `storage_system` rows. Is this desirable? Also, earlier we discussed using a composite primary key instead (but not a string). At the UI layer, will we use these `uuid`s? Or will there be a translation from `tdr_<uuid>` to this `uuid`? I think it would be convenient if we encode some additional information in `uuid`s (probably not at the database layer), that way users at least know they're using the right `uuid` for their systems, and we detect some simple things with this too ("oh, you're trying to add a TDR id to the wrong storage system"). GitHub for example does this for personal access tokens: they're opaque blobs that have a `ghp_` prefix, which enables them to scan for secrets accidentally being committed in repositories.

fboulnois (commented):
A few questions:

service/src/main/resources/db/changesets/20220301_datasettable.yaml
samanehsan: The unique constraint will ensure that the combination of `dataset_id` and `storage_system` will be unique (I added some unit tests to check this behavior). 

I was thinking as we eventually support different storage systems, it might confuse users to see ids in multiple different formats. For instance, what would the id for an "external" dataset look like (where we just provide a link to the dataset)? We could still validate that the dataset_id is in a valid format for the storage system when a user tries to create a catalog entry. What do you think?

common/postgres-init.sql
```diff
@@ -1,2 +1,3 @@
 CREATE DATABASE catalog_db;
 CREATE ROLE dbuser WITH LOGIN ENCRYPTED PASSWORD 'dbpwd';
+CREATE EXTENSION pgcrypto;
```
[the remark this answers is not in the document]
samanehsan: I tested this for my local database but I need to look into how to do this for our deployed databases

service/src/main/resources/db/changesets/20220301_datasettable.yaml
```diff
@@ -0,0 +1,40 @@
...
+                  type: jsonb
+                  constraints:
+                    nullable: true
+              - column:
+                  name: created_date
+                  type: timestamptz
+                  defaultValueComputed: now()
+                  constraints:
+                    nullable: false
+        - addUniqueConstraint:
```
fboulnois, line 37:
```suggestion
        # ensure there can only be one dataset_id, storage_system combination
        - addUniqueConstraint:
```

fboulnois: I missed this at first, maybe this comment is redundant though 👍

fboulnois (approved):
I understand the design better now, I think it's worth thinking about the `uuid` issue, but that shouldn't block this PR :)

service/src/main/resources/db/changesets/20220301_datasettable.yaml
samanehsan: I'll add it! It's not really clear from just looking at the yaml if each field needs to be unique or the combination of the two.

petesantos (commented):
LGTM. I'll let Phil give it the final thumb.

samanehsan:
I updated sonar to ignore the error about having too many parent classes but the test coverage report still fails

service/src/main/java/bio/terra/catalog/common/DaoKeyHolder.java
```diff
@@ -0,0 +1,42 @@
+package bio.terra.catalog.common;
```
lu-c, line 1: Are these files generated? From the swagger stuff?  Just curious, not asking for change

service/src/main/resources/db/changelog.xml
```diff
@@ -1,3 +1,7 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+  <property name="uuid_type" value="uuid" dbms="postgresql"/>
```
lu-c, line 3: How did you know what files here to modify? (again, no actions necessary, just tryign to understand the backend a bit more)

service/src/main/java/bio/terra/catalog/common/DaoKeyHolder.java
pshapiro4broad: There are many files generated by swagger, but we don't check them in. The swagger task will regenerate the files as prerequisite to the java compile task if necessary.

service/src/main/java/bio/terra/catalog/common/DaoKeyHolder.java
```diff
@@ -0,0 +1,42 @@
...
+
+  public String getString(String fieldName) {
+    return getField(fieldName, String.class).orElse(null);
+  }
+
+  public <T> Optional<T> getField(String fieldName, Class<T> type) {
+    Map<String, Object> keys = getKeys();
+    if (keys != null) {
+      Object fieldObject = keys.get(fieldName);
+      if (type.isInstance(fieldObject)) {
```
pshapiro4broad, line 36: As far as I can tell, this is the main useful line in this method. Under what circumstances are we getting back an incorrectly typed object? In those cases do we want to ignore the type difference and return `null` instead? Unless we have a clear case where we know there's a type mismatch, I think it would be better to return `keys.get(fieldName)` cast to the expected type directly. A type mismatch will result in a thrown runtime exception, but I think we'd want to know about this and not hide the error.

service/src/main/java/bio/terra/catalog/service/dataset/DatasetDao.java
```diff
@@ -0,0 +1,120 @@
...
+import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
+import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
+import org.springframework.stereotype.Repository;
+
+@Repository
+public class DatasetDao {
+
+  private final NamedParameterJdbcTemplate jdbcTemplate;
+  private static final String ID_FIELD = "id";
+  private static final String DATASET_ID_FIELD = "dataset_id";
```
pshapiro4broad, line 26: The more I see this, the more I think we should have a different name for this field. If I was writing a comment in the code and was talking about "dataset id", which ID would I be talking about? I think the "id" field on "dataset" would be the "dataset id". Maybe this should be "storage ID" (or "resource ID")?

service/src/test/java/bio/terra/catalog/service/dataset/DatasetDaoTest.java
```diff
@@ -0,0 +1,72 @@
...
+  private Dataset createDataset(String datasetId, StorageSystem storageSystem, String metadata)
+      throws InvalidDatasetException {
+    Dataset dataset = new Dataset(null, datasetId, storageSystem, metadata, null);
+    return datasetDao.create(dataset);
+  }
+
+  @Test
+  void testDatasetCrudOperations() {
+    String datasetId = UUID.randomUUID().toString();
+    String metadata = "{\"sampleId\": \"12345\", \"species\": [\"mouse\", \"human\"]}";
```
pshapiro4broad, line 35: It looks like this is the same for all tests; if so, I would make it a static final field in the test class. Also JDK 17 includes string blocks, so you could write this as:
```suggestion
    String metadata = """
                                  {"sampleId": "12345", "species": ["mouse", "human"]}""";
```

pshapiro4broad (approved):
Looks good, just a few more comments / suggestions.

service/src/main/resources/db/changelog.xml
samanehsan: We're using [liquibase](https://liquibase.org/) to manage changes to our postgres database. To make a new change, I created a "changeset" file specifying the operations I wanted to do. Liquibase will apply all of the changesets in `changelog.xml` and create a db table to track what was already run.

service/src/main/java/bio/terra/catalog/common/DaoKeyHolder.java
samanehsan: I borrowed this from the data repo, but that's a good point that we don't want to hide the error!

PR #23, opened 2022-03-03
[Experiment] Test spring data jdbc

samanehsan:
Test out `spring-data-jdbc` to simplify database operations: https://spring.io/blog/2018/09/17/introducing-spring-data-jdbc

The test is failing because it can't load the context 😞


samanehsan:
I'll leave this branch around in case we want to revisit this later!

PR #25, opened 2022-03-03
DC-251: Include GitHub token

fboulnois:
I think this should fix the GitHub action issue with SonarQube


pshapiro4broad (approved):
This seems weird to me. If it's needed, why did the sonarqube scan on my PR succeed? The error message mentions the GitHub action, but we aren't using the action, as far as I can tell.

samanehsan (approved):

PR #27, opened 2022-03-04
DC-253: Run unit tests on PRs

pshapiro4broad:
Add the `test` task to the jacoco gradle target so unit tests are run.

- generate an XML jacoco report for sonarqube
- run unit-tests before sonarqube so XML report is present (not working yet)
- use fetch-depth to checkout target merge branch for sonarqube

The ticket https://broadworkbench.atlassian.net/browse/DC-262 tracks the remaining work to upload coverage data using sonarqube.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -128,7 +128,7 @@ jobs:
           cache: 'gradle'
 
       - name: SonarQube scan
-        run: ./gradlew --build-cache -Dsonar.login="${{ secrets.SONAR_TOKEN }}" sonarqube --info
+        run: ./gradlew --build-cache -Dsonar.login="${{ secrets.SONAR_TOKEN }}" test jacocoTestReport sonarqube --info
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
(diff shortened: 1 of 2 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -183,7 +181,7 @@ jobs:
         run: ./gradlew --build-cache :integration:runTest --args="suites/FullIntegration.json /tmp/foo"
 
   notify-slack:
-    needs: [ build, jib, unit-tests, source-clear, sonar-scanner, integration-tests ]
+    needs: [ build, jib, source-clear, sonar-scanner, integration-tests ]
```
fboulnois, line 184: I think it's fine to keep `unit-tests` here, that way if we move stuff around we don't need to manually figure out the dependency tree (otherwise for consistency should we also remove `build`?)

fboulnois (approved):
One suggestion:

.github/workflows/build-and-test.yml
pshapiro4broad: Good point, I'll add it back

samanehsan (approved):
Lgtm! Do we have another ticket for fixing the coverage report?

PR #31, opened 2022-03-08
DC-150: Add Apache v2.0 license

fboulnois:


pshapiro4broad (approved):

samanehsan (approved):

pshapiro4broad:
Do you need to do anything special to get the license banner like the one shown in https://github.com/DataBiosphere/jade-data-repo/blob/develop/LICENSE.md ?

I would expect to see something like that when viewing the license file here: https://github.com/DataBiosphere/terra-data-catalog/blob/fb-dc-150-add-license/LICENSE

fboulnois:
> Do you need to do anything special to get the license banner like the one shown in https://github.com/DataBiosphere/jade-data-repo/blob/develop/LICENSE.md ?
> 
> I would expect to see something like that when viewing the license file here: https://github.com/DataBiosphere/terra-data-catalog/blob/fb-dc-150-add-license/LICENSE

I did it by adding the LICENSE the official way, so it should appear when it's merged.

PR #33, opened 2022-03-11
[DC-194] Add liquibase gradle plugin

samanehsan:
Configure the liquibase gradle plugin to make it easier to run database migrations. The commands can be run with ./gradlew :service:<liquibase_command>, e.g. ./gradlew :service:update (although if there's a way to include a liquibase prefix here that would make it clearer)

Note: versions 4.0.0 and above have an [issue](https://github.com/liquibase/liquibase/issues/2281) with including changeset file paths relative to the changelog file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,3 +45,20 @@ To run performance tests, execute:
 render-configs.sh perf
 ./gradlew :integration:runTest --args="suites/FullPerf.json /tmp/test"
 ```
+
+## Handling Database Migrations
+The catalog service uses [Liquibase](https://liquibase.org/) to track and manage changes to the
+database schema. Liquibase runs each changeset (migration) listed in the
+[changelog.xml](service/src/resources/db/changelog.xml) file and
+maintains a record of what has been run, so new changes must be added in a new changeset.
+
+To run migrations locally use:
+```
+./gradlew update
+```
+
+If the local database gets into a bad state (for instance while testing/modifying a new changeset),
+drop its contents with:
+```
+./gradlew dropAll
+```
--- a/README.md
+++ b/README.md
@@ -46,7 +46,8 @@ render-configs.sh perf
 ./gradlew :integration:runTest --args="suites/FullPerf.json /tmp/test"
 ```
 
-## Handling Database Migrations
+## Handling database migrations
+
 The catalog service uses [Liquibase](https://liquibase.org/) to track and manage changes to the
 database schema. Liquibase runs each changeset (migration) listed in the
 [changelog.xml](service/src/resources/db/changelog.xml) file and
--- a/README.md
+++ b/README.md
@@ -53,6 +53,7 @@ database schema. Liquibase runs each changeset (migration) listed in the
 maintains a record of what has been run, so new changes must be added in a new changeset.
 
 To run migrations locally use:
+
 ```
 ./gradlew update
 ```
--- a/README.md
+++ b/README.md
@@ -61,6 +61,7 @@ To run migrations locally use:
 
 If the local database gets into a bad state (for instance while testing/modifying a new changeset),
 drop its contents with:
+
 ```
 ./gradlew dropAll
 ```
--- a/service/build.gradle
+++ b/service/build.gradle
@@ -48,7 +48,7 @@ liquibase {
     activities {
         catalog {
             changeLogFile 'src/main/resources/db/changelog.xml'
-            url 'jdbc:postgresql://127.0.0.1:5432/catalog_db'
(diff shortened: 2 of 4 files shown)


service/gradle.properties
```diff
@@ -1,6 +1,13 @@
...
-org.gradle.jvmargs=--add-exports jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED \
+org.gradle.jvmargs= \
+  --add-exports jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED \
   --add-exports jdk.compiler/com.sun.tools.javac.file=ALL-UNNAMED \
   --add-exports jdk.compiler/com.sun.tools.javac.parser=ALL-UNNAMED \
   --add-exports jdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED \
   --add-exports jdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED
+
+# Liquibase properties
+dbUri=jdbc:postgresql://127.0.0.1:5432/catalog_db
```
pshapiro4broad, line 10: Is there a reason to store the properties here instead of putting them the `build.gradle` file? I think it would be better to only have them in the `build.gradle` unless there's a benefit to using a properties file too

pshapiro4broad (commented):
> The commands can be run with ./gradlew :service:<liquibase_command>, e.g. ./gradlew :service:update (although if there's a way to include a liquibase prefix here that would make it clearer)

In general you should be able to run a subproject task without having to specify the subproject, unless you're trying to run the task in only a specific subproject (e.g., if multiple subproject have a `test` task and you only want to run it in `service`). In cases like this I think it's easier to document and use if you omit the subproject. Using `./gradlew dropAll` should work fine.

Can you add a couple lines to the readme that mention a couple common use cases for this plugin? Something like

```
If you're debugging a liquibase migration, you can iterate on it by using `update` to run the migration. If your local catalog database gets into a bad state, you can reset it by using `dropAll` to remove all entities.
```

service/gradle.properties
samanehsan: No particular reason! I thought it might be clearer to have that separation but it also seems to break things

pshapiro4broad (approved):
Looks good!

service/build.gradle
```diff
@@ -37,3 +43,15 @@ sonarqube {
         property 'sonar.host.url', 'https://sonarcloud.io'
     }
 }
+
+liquibase {
+    activities {
+        catalog {
+            changeLogFile 'src/main/resources/db/changelog.xml'
+            url 'jdbc:postgresql://127.0.0.1:5432/catalog_db'
```
fboulnois, line 51:
```suggestion
            url 'jdbc:postgresql://localhost:5432/catalog_db'
```

I think `localhost` is more portable: on MacOS/Linux this is translated to a local socket connection which can be faster than TCP/IP and if you theoretically debug on an IPV6 only system `localhost` should just work.

README.md
```diff
@@ -45,3 +45,20 @@ To run performance tests, execute:
...
 ./gradlew :integration:runTest --args="suites/FullPerf.json /tmp/test"
 ```
+
+## Handling Database Migrations
+The catalog service uses [Liquibase](https://liquibase.org/) to track and manage changes to the
+database schema. Liquibase runs each changeset (migration) listed in the
+[changelog.xml](service/src/resources/db/changelog.xml) file and
+maintains a record of what has been run, so new changes must be added in a new changeset.
+
+To run migrations locally use:
```
fboulnois, line 55:
```suggestion
To run migrations locally use:

```

Nit: add newline?

README.md
```diff
@@ -45,3 +45,20 @@ To run performance tests, execute:
...
+```
+./gradlew update
+```
+
+If the local database gets into a bad state (for instance while testing/modifying a new changeset),
+drop its contents with:
```
fboulnois, line 61:
```suggestion
drop its contents with:

```

Nit: add newline?

README.md
```diff
@@ -45,3 +45,20 @@ To run performance tests, execute:
 render-configs.sh perf
 ./gradlew :integration:runTest --args="suites/FullPerf.json /tmp/test"
 ```
+
+## Handling Database Migrations
```
fboulnois, line 49:
```suggestion
## Handling database migrations

```

Nit: capitalization and add newline?

fboulnois (approved):
Minor suggestions:

service/build.gradle
samanehsan: That's good to know!

PR #34, opened 2022-03-14
Update postgres-init.sql

samanehsan:
Suggested by @pshapiro4broad: switch to the `catalog_db` to guarantee the correct database is being modified.

### DIFF
--- a/common/postgres-init.sql
+++ b/common/postgres-init.sql
@@ -1,3 +1,4 @@
 CREATE DATABASE catalog_db;
+\c catalog_db;
 CREATE ROLE dbuser WITH LOGIN ENCRYPTED PASSWORD 'dbpwd';
 CREATE EXTENSION pgcrypto;


common/postgres-init.sql
```diff
@@ -1,3 +1,4 @@
 CREATE DATABASE catalog_db;
+\c catalog_db;
```
pshapiro4broad, line 2: since this is a psql command and not sql it doesn't need a `;`
```suggestion
\c catalog_db
```

pshapiro4broad (approved):
One minor comment, thanks for updating this

common/postgres-init.sql
```diff
@@ -1,3 +1,4 @@
 CREATE DATABASE catalog_db;
+\c catalog_db
 CREATE ROLE dbuser WITH LOGIN ENCRYPTED PASSWORD 'dbpwd';
 CREATE EXTENSION pgcrypto;
```
fboulnois, line 4:
```suggestion
CREATE ROLE dbuser WITH LOGIN ENCRYPTED PASSWORD 'dbpwd';
CREATE DATABASE catalog_db OWNER dbuser;
\c catalog_db
CREATE EXTENSION pgcrypto;
```

While we're fixing stuff in `postgres-init.sql`, I'm curious if instead we might want the `dbuser` to own the `catalog_db`?

fboulnois (approved):
Can you remind me why we need the `pgcrypto` extension? From what I remember it is to generate random `uuid`s, however as of version 13 of Postgres, the function [is built-in](https://www.postgresql.org/docs/13/functions-uuid.html). Also, do we want `dbuser` to own the `catalog_db`?

common/postgres-init.sql
samanehsan: That makes sense to me!

samanehsan:
@fboulnois I'm leaning towards leaving the pgcrypto extension in this script for backwards-compatibility. This script only needs to be used to set up a local database, since terraform handles creating a database user and password for our deployments. I'm not sure if that user is created as an owner by default though.

PR #35, opened 2022-03-15
DC-126: Report nightly tests to dsde-qa

fboulnois:
For compliance reasons we need to support both success and failures when we run perf tests to dsde-qa.

### DIFF
--- a/.github/workflows/nightly-tests.yml
+++ b/.github/workflows/nightly-tests.yml
@@ -81,14 +81,14 @@ jobs:
           name: Test Reports
           path: integration/build/reports
 
-  notify-slack:
+  notify-slack-failure:
     needs: [ build, jib, test-runner ]
     runs-on: ubuntu-latest
 
     if: failure() && github.ref == 'refs/heads/main'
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly-tests.yml
```diff
@@ -81,14 +81,14 @@ jobs:
           name: Test Reports
           path: integration/build/reports
 
-  notify-slack:
+  notify-slack-failure:
     needs: [ build, jib, test-runner ]
     runs-on: ubuntu-latest
 
     if: failure() && github.ref == 'refs/heads/main'
```
pshapiro4broad, line 88: I was going to comment that the new `if` below needed the `github.ref == 'refs/heads/main'` check, but now I think the reverse is true, that the check here isn't needed, as the only trigger for this workflow is on the `main` branch.

fboulnois: Yes I think that's right. The workflow is only triggered on the `main` branch, so the condition will always be true.

PR #37, opened 2022-03-20
DC-278: Use a session scoped bean for the authenticated user

pshapiro4broad:
Based on a suggestion from @petesantos, this change uses a http request scoped bean to generate the current authenticated user's token. The notable improvement of this approach is that the code no longer needs to pass the request's user all the way from the controller down to the service layer. With this change, the `User` is only injected in the components that needs it, which currently are `DatarepoClient`, `RawlsClient` and `SamService`.

This also removes the code that used SAM to retrieve the current user's email. Our code doesn't use the email, and we shouldn't be using SAM for this purpose anyway; in general we shouldn't call SAM on every incoming request unless it's necessary. If we need the user email, we should be extracting it from the OIDC proxy generated headers.


service/src/main/java/bio/terra/catalog/api/DatasetApiController.java
```diff
@@ -6,62 +6,46 @@
 import bio.terra.catalog.model.DatasetsListResponse;
```
fboulnois, line 6: test

s-rubenstein (commented):
I'd love a test to make sure the cache doesn't bleed between requests. I assume it doesn't, but whenever user credentials are getting cached I have a little worry about things getting messed up.

pshapiro4broad:
> I'd love a test to make sure the cache doesn't bleed between requests. I assume it doesn't, but whenever user credentials are getting cached I have a little worry about things getting messed up.

There's probably a way to write a mock MVC test that simulates different sessions, let me think about how to do this.

service/src/main/java/bio/terra/catalog/iam/SamAuthenticatedUserRequestFactory.java
```diff
@@ -2,33 +2,36 @@
...
-  private static final String OAUTH2_ACCESS_TOKEN = "OAUTH2_CLAIM_access_token";
-  private static final String AUTHORIZATION = "Authorization";
-  private final SamService samService;
+@SessionScope
+public class SamAuthenticatedUserRequestFactory {
+  static final String OAUTH2_ACCESS_TOKEN = "OAUTH2_CLAIM_access_token";
+  static final String AUTHORIZATION = "Authorization";
+  private final SamUserStatusService samService;
+  private final HttpServletRequest request;
+  private AuthenticatedUserRequest cachedUser;
```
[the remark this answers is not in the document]
pshapiro4broad: > What expires the cache/when does the user expire? And how is it cached specifically, is it based on accessToken? Or request context?

The lifetime is tied to the containing object (`SamAuthenticatedUserRequestFactory`) so it's is determined by the scope ( of that object. Since it has `@SessionScope`, its lifetime is tied to the user's browser session. I think session is the correct scope to use, but the other HTTP scope to use is `RequestScope`. That won't cache the value across multiple requests, but will cache it within a single HTTP request operation.

If you're not familiar with bean scopes, check out https://docs.spring.io/spring-framework/docs/current-SNAPSHOT/reference/html/core.html#beans-factory-scopes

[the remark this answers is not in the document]
pshapiro4broad: I've changed the code to use request scope instead, since our app doesn't have sessions. This is "safe" in that spring will throw an error if a component tries to use a request scope object outside of a request scope. Although right now everything our app does is only through request scoped operations.

pshapiro4broad:
> > I'd love a test to make sure the cache doesn't bleed between requests. I assume it doesn't, but whenever user credentials are getting cached I have a little worry about things getting messed up.
>
> There's probably a way to write a mock MVC test that simulates different sessions, let me think about how to do this.

I started to write some tests but found that I'd be pretty much only testing my mocks, which didn't seem very useful. We do have integration tests which perform permission operations, and if there was a way for a user to "leak" between requests then I think the tests would fail.

service/src/main/java/bio/terra/catalog/iam/User.java
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * This represents the user associated with the current HTTP request. If injected into a singleton
+ * scope class, such as a service, spring will create a proxy object so the value for the current
+ * http request is returned. If no http scope exists when this is used, the proxy will throw an
+ * exception.
+ */
+@Component
+@RequestScope
+public class User {
```
pshapiro4broad, line 18: Assuming https://github.com/DataBiosphere/terra-data-catalog/pull/133 is merged, this can probably be replaced with `BearerToken`.

service/src/main/java/bio/terra/catalog/iam/User.java
```diff
@@ -0,0 +1,45 @@
...
+    String oauth2Header = httpRequest.getHeader(OAUTH2_ACCESS_TOKEN);
+    if (oauth2Header != null) {
+      return oauth2Header;
+    }
+    String authHeader = httpRequest.getHeader(AUTHORIZATION);
+    if (authHeader != null) {
+      // Invalid tokens will throw an UnauthorizedException.
+      return BearerTokenParser.parse(authHeader);
+    }
+    throw new UnauthorizedException("Unable to retrieve access token");
```
s-rubenstein, line 43: OPT: Since we are messing around with access tokens and exceptions, do we want to change the other UnauthorizedException returned by the access check to be a ForbiddenException?

service/src/test/java/bio/terra/catalog/datarepo/DatarepoClientTest.java
```diff
@@ -2,31 +2,39 @@
...
   void testApis() {
-    var user =
-        new AuthenticatedUserRequest.Builder()
-            .setEmail("")
-            .setSubjectId("")
-            .setToken(TOKEN)
-            .build();
-
-    var snapshotsClient = client.snapshotsApi(user).getApiClient();
+    when(user.getToken()).thenReturn(TOKEN);
```
s-rubenstein, line 35: This is much cleaner of a mock!

service/src/test/java/bio/terra/catalog/service/DatasetServiceTest.java
```diff
@@ -177,71 +178,69 @@ void listDatasetsUsingAdminPermissions() {
...
-    reset(datasetDao);
-    when(datasetDao.retrieve(datasetId)).thenReturn(tdrDataset);
-    when(datarepoService.getRole(user, sourceId)).thenReturn(DatasetAccessLevel.READER);
-    datasetService.getMetadata(user, datasetId);
-    verify(datasetDao).retrieve(datasetId);
+    when(datasetDao.retrieve(tdrDataset.id())).thenReturn(tdrDataset);
+    when(datarepoService.getRole(sourceId)).thenReturn(DatasetAccessLevel.READER);
+    assertThat(
+        datasetService.getMetadata(tdrDataset.id()),
+        is(METADATA_WITH_ID.formatted(tdrDataset.id().uuid())));
```
s-rubenstein, line 212: Optional Nit: I personally prefer working with object structures rather than formatted strings, because then order doesn't matter for things like `.formatted`. Would it be possible to have `METADATA_WITH_ID` be an object, and hold off stringifying it until we get to this point here?

service/src/main/java/bio/terra/catalog/iam/User.java
pshapiro4broad: I've replaced it with `BearerToken` and moved the bean wiring to `BeanConfig`.

service/src/test/java/bio/terra/catalog/iam/UserTest.java
```diff
@@ -0,0 +1,38 @@
...
+class UserTest {
+
+  private final MockHttpServletRequest request = new MockHttpServletRequest();
+
+  private final User user = new User(request);
+
+  @Test
+  void getTokenOauth2() {
+    var token = "token";
+    request.addHeader(User.OAUTH2_ACCESS_TOKEN, token);
```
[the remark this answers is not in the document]
pshapiro4broad: This code is now gone, but to answer the question, there's no risk because JUnit creates a new class instance for each test method invocation. So each test method gets a different copy of the test class. (I seem to recall TestNG shares the test class instance across test methods, so maybe that's what you're thinking about here.)

service/src/main/java/bio/terra/catalog/iam/User.java
pshapiro4broad: This code is now gone, as the new code is calling `BearerTokenFactory` in the tcl, which handles this case. I think fixing the forbidden case (which looks like it would be a change to `DatasetService.ensureActionPermission()`) should be done in a separate PR. I created https://broadworkbench.atlassian.net/browse/DC-574 for this issue.

service/src/test/java/bio/terra/catalog/rawls/RawlsClientTest.java
```diff
@@ -2,34 +2,41 @@
...
   void testApis() {
-    var user =
-        new AuthenticatedUserRequest.Builder()
-            .setEmail("")
-            .setSubjectId("")
-            .setToken(TOKEN)
-            .build();
-
-    ApiClient workspacesClient = client.workspacesApi(user).getApiClient();
+    when(user.getToken()).thenReturn(TOKEN);
```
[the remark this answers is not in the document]
pshapiro4broad: Good idea, I've moved it.

s-rubenstein (approved):

Abulu123 (approved):
Looks good to me!

PR #38, opened 2022-03-21
[DC-218] Added two data catalog test user accounts

lu-c:
Copied the structure that existed for Scarlett Flowerpicker - Just updated the userEmail field. Please let me know if I'm missing something!

### DIFF
--- a/integration/src/main/resources/configs/integration/ListDatasets.json
+++ b/integration/src/main/resources/configs/integration/ListDatasets.json
@@ -13,5 +13,5 @@
       "expectedTimeForEachUnit": "SECONDS"
     }
   ],
-  "testUserFiles": [ "scarlett.json" ]
+  "testUserFiles": [ "user.json", "admin.json" ]
 }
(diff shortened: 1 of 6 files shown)


samanehsan (approved):
Lgtm!

fboulnois (approved):

pshapiro4broad (changes requested):
Can you update the integration test configuration to replace `scarlett.json` with `user.json`? If this user exists in dev terra, then the test should continue to work. Then you can also delete the `scarlett.json` file.

integration/src/main/resources/configs/integration/ListDatasets.json
```diff
@@ -13,5 +13,5 @@
       "expectedTimeForEachUnit": "SECONDS"
     }
   ],
-  "testUserFiles": [ "scarlett.json" ]
+  "testUserFiles": [ "user.json", "admin.json" ]
```
pshapiro4broad, line 16: I think doing this will cause the test to be run multiple times, once per test user. At this point I don't think there's a benefit to doing this, since the behavior will be the same for both users.

pshapiro4broad (approved):
Look good, just one more comment/question about users in tests.

integration/src/main/resources/configs/integration/ListDatasets.json
lu-c: I can remove the admin, np! I usually like to test everything cause the only sacrifice we make is time, and ... ya never know!

PR #40, opened 2022-03-28
DC-285: Notify Slack in a single job

fboulnois:
The reported job status is inconsistent between jobs. When the job fails, the GitHub Action reports that nightly test failed in `jade-data-explorer` but reports success in `dsde-qa`. Put them in the same step to try to fix the issue.

### DIFF
--- a/.github/workflows/nightly-tests.yml
+++ b/.github/workflows/nightly-tests.yml
@@ -81,15 +81,13 @@ jobs:
           name: Test Reports
           path: integration/build/reports
 
-  notify-slack-failure:
+  notify-slack:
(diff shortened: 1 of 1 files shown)


lu-c:
The code looks fine to me - Have you tried forcing an error and seeing if this change works?

fboulnois:
> The code looks fine to me - Have you tried forcing an error and seeing if this change works?

Yep good idea, I'll trigger a workflow dispatch.

pshapiro4broad (approved):

lu-c (approved):

PR #41, opened 2022-03-28
DC-270: add list datasets support

pshapiro4broad:


fboulnois (approved):
One minor suggestion:

integration/src/main/resources/suites/FullIntegration.json
```diff
@@ -4,7 +4,6 @@
   "serverSpecificationFile": "catalog-local.json",
   "testConfigurationFiles": [
     "integration/GetStatus.json",
-    "integration/GetVersion.json",
-    "integration/ListDatasets.json"
```
lu-c, line 8: Are we going to have tests for the ListDatasets function?

lu-c (approved):
LGTM

PR #43, opened 2022-03-29
[DC-231] Add swagger-ui.html support

lu-c:
implement a swagger-ui.html file that will take `/common/openapi.yml` and create a swagger ui page.

- Added task `copyOpenApiToResources` to the gradle services
-

Created this PR for stephen lucas, please refer to him with your suggestions and comments


README.md
```diff
@@ -40,7 +40,7 @@ render-configs.sh      # render service account credentials needed for tests
 ```
 
 To run performance tests, execute:
-
+s
```
lu-c, line 43: Should this be here?

pshapiro4broad, line 43: typo
```suggestion
```

service/generators.gradle
```diff
@@ -34,6 +34,14 @@ idea.module.generatedSourceDirs = [file(swaggerOutputSrc)]
 sourceSets.main.java.srcDir swaggerOutputSrc
 compileJava.dependsOn generateSwaggerCode
 
+// Copy the result of the openapi build into the built resources
+// so it is available for the swagger_ui.html endpoint
+task copyOpenApiToResources(type: Copy) {
+    description 'Copy the openapi results into build resources to support swagger_ui'
+    from '../common/openapi.yml'
```
pshapiro4broad, line 41: It looks like we've now repeated `../common/openapi.yml` in three different places. It would be great if it could be replaced with a shared variable. I think defining it in `bio.terra.catalog.java-common-conventions.gradle` would work, the same way we do for `version` and `group`. Also using a relative path here is brittle, if something like `"${buildDir}/common/openapi.yml"` works that would be better.

service/src/main/java/bio/terra/catalog/app/App.java
```diff
@@ -27,6 +29,8 @@
       // Scan all service-specific packages beneath the current package
       "bio.terra.catalog"
     })
+@OpenAPIDefinition(info = @Info(title = "Terra", version = "2.0"))
```
pshapiro4broad, line 32: Is this needed? I don't see it used in the jade-data-repo sources.

service/src/main/java/bio/terra/catalog/config/BaseDatabaseConfiguration.java
```diff
@@ -8,6 +8,7 @@
 import org.apache.commons.dbcp2.PoolingDataSource;
 import org.apache.commons.pool2.ObjectPool;
 import org.apache.commons.pool2.impl.GenericObjectPool;
+import org.springframework.web.bind.annotation.RestController;
```
pshapiro4broad, line 11: unused import?

service/src/main/resources/api/swagger-ui.html
```diff
@@ -0,0 +1,37 @@
+<!-- HTML for static distribution bundle build -->
```
pshapiro4broad, line 1: I'd be interested to see if we could use a similar setup to what jade-data-repo is doing here. Or maybe that can wait for a future PR. Specifically they have a template version of this file that substitutes `clientId` and `clientSecret` when the page is rendered. I think we need something similar to allow the swagger page to be used for authenticated requests. See https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/resources/templates/index.html

service/src/main/java/bio/terra/catalog/app/App.java
slucasbroad: it was left over from me experimenting with learning notations. I will remove it if thats ok?

service/src/main/java/bio/terra/catalog/config/BaseDatabaseConfiguration.java
slucasbroad: Got rid of it in most recent version. Probably forgot to list the file when updating branch.

service/generators.gradle
fboulnois: We can also commit a symlink for each of these locations, that seems to be what `external-credentials-manager` does: https://github.com/DataBiosphere/terra-external-credentials-manager/search?q=openapi

PR #44, opened 2022-03-30
DC-158: Add warning in OpenApi description

fboulnois:
To meet the requirements for SC11 - Authorization Domains, we need to include a warning that all entries will be public to Terra users.

### DIFF
--- a/common/openapi.yml
+++ b/common/openapi.yml
@@ -52,7 +52,10 @@ paths:
           $ref: '#/components/responses/ServerError'
     post:
       tags: [ datasets ]
-      description: Create a new dataset
+      description: |
+        ⚠️ Do not add sensitive data. All catalog entries are publicly accessible to all Terra users ⚠️
+      
+        Create a new dataset
       operationId: createDataset
       requestBody:
         content:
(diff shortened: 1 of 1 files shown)


common/openapi.yml
```diff
@@ -52,7 +52,10 @@ paths:
           $ref: '#/components/responses/ServerError'
     post:
       tags: [ datasets ]
-      description: Create a new dataset
+      description: |
```
pshapiro4broad, line 55: Did you mean to replace the `Create a new dataset` text here? Or maybe that text should go in a `summary` property instead of in `description`?

common/openapi.yml
```diff
@@ -52,7 +52,10 @@ paths:
...
+        ⚠️ Do not add sensitive data. All catalog entries are publicly accessible to all Terra users ⚠️
```
pshapiro4broad, line 56: Maybe this should be a note on the `metadata` property in `CreateDatasetRequest` instead of a general warning on the API, since the warning applies to that property specifically.

fboulnois: I think we want to make it as visible as possible, so we should add it to the main page and on the `CreateDatasetRequest` object too.

common/openapi.yml
fboulnois: The text is still in the description (see below), but I think for consistency we should switch these descriptions to the summary section instead.

pshapiro4broad (approved):

petesantos (approved):

PR #50, opened 2022-04-05
[DC-221] adding metadata to datasets?!

lu-c:
Maybe this should be another draft pr! Dunno!

Adding metadata to datasets in the list that are identified as testing datasets:

High level pseudocode:
- list datasets
- find matching metadata in resources.json that are keyed by dataset[dct:title]

Current Problems:
- calls list datasets, then tries to PUT via the ids acquired from the request, but gets this problem:
```Updating metadata for datasets in the catalog
datasetList [{'id': '1568e68d-3032-4571-a066-dac6ccec9716', 'dct:title': 'DiscoverableCatalogSnapshot1'}, ...]

Updating dataset (1568e68d-3032-4571-a066-dac6ccec9716, DiscoverableCatalogSnapshot1)
---------------------------------------------------
There was a problem updating metadata for (1568e68d-3032-4571-a066-dac6ccec9716, DiscoverableCatalogSnapshot1)
Dataset not found for DatasetId[uuid=1568e68d-3032-4571-a066-dac6ccec9716]
```

Might still have permission problems as well? But until this first hurdle is leapt, not sure if it's been dealt with


scripts/catalog-tests/generate-tdr-snapshot-metadata.py
```diff
@@ -0,0 +1,85 @@
+#------------------------------------------------------------------------------
+# Generate TDR Snapshot Metadata
+#
+# This script adds snapshots to the catalog as the test user from DC-218,
+# pulling the metadata from the resources JSON file
+# 
+# Before Running:
+#   python3 -m pip install requests argparse
```
fboulnois, line 8:
```suggestion
#   python3 -m venv .env
#   source .env/bin/activate
#   pip install --upgrade pip
#   python3 -m pip install requests argparse
```

I suggest installing all the package in a virtualenv, `pip install` is brittle and can break your Python installation which can be difficult to fix on MacOS :(

scripts/catalog-tests/generate-tdr-snapshot-metadata.py
```diff
@@ -0,0 +1,85 @@
...
+#   
+#   Authorization: Access the password in the vault using the appropriate command:
+#   vault read secret/dsde/terra/kernel/dev/dev/catalog/tests/user
```
fboulnois, line 11: You might need to run `vault login -method=github token=$(cat ~/.github-token)` or equivalent beforehand to authenticate.

scripts/catalog-tests/generate-tdr-snapshot-metadata.py
```diff
@@ -0,0 +1,85 @@
+#------------------------------------------------------------------------------
+# Generate TDR Snapshot Metadata
+#
+# This script adds snapshots to the catalog as the test user from DC-218,
+# pulling the metadata from the resources JSON file
+# 
+# Before Running:
+#   python3 -m pip install requests argparse
+#   
+#   Authorization: Access the password in the vault using the appropriate command:
```
fboulnois, line 10: How is the vault password used in the script?

fboulnois (commented):
A few suggestions:

scripts/catalog-tests/generate-tdr-snapshot-metadata.py
lu-c: It's not, unfortunately - I couldn't figure out how to do it in a way that would be completely contained in the script. The commands to gcloud will open a browser window and prompt you to log in to that user.

scripts/catalog-tests/generate-tdr-snapshot-metadata.py
lu-c: ok! :)

scripts/catalog-tests/generate-tdr-snapshot-metadata.py
lu-c: Added!

scripts/catalog-tests/generate-tdr-snapshot-metadata.py
fboulnois: Oh I see. This is fine 👍

fboulnois (commented):
Another thought, perhaps we can format everything with `black`? That way all the Python scripts are formatted consistently.

scripts/catalog-tests/generate-tdr-snapshot-metadata.py
```diff
@@ -0,0 +1,127 @@
...
+# Before Running:
+#   python3 -m venv .env
+#   source .env/bin/activate
+#   pip install --upgrade pip
+#   python3 -m pip install requests
+#   
+#   Authorization: Access the password in the vault using the appropriate command:
+#   vault login -method=github token=$(cat ~/.github-token)
+#   vault read secret/dsde/terra/kernel/dev/dev/catalog/tests/user
+#   vault read secret/dsde/terra/kernel/dev/dev/catalog/tests/userAdmin
```
pshapiro4broad, line 16: Could you add a note to make it clear that the password is needed for the auth dialog that the script will put up? Also this prints both passwords but I think only `userAdmin` is needed.

Also maybe add a note on how the env is used here. If you're running this for another environment, you'd have to change a few paths in this script.

scripts/catalog-tests/generate-tdr-snapshot-metadata.py
```diff
@@ -0,0 +1,127 @@
...
+# Run: python3 generate-tdr-snapshot-metadata.py
```
pshapiro4broad, line 17: Maybe give an example using `catalogServiceUrl` so the user knows how to configure the tool.

pshapiro4broad, line 17: Maybe give an example using `catalogServiceUrl` so the user knows how to configure the tool.

scripts/catalog-tests/generate-tdr-snapshot-metadata.py
```diff
@@ -0,0 +1,127 @@
...
+#------------------------------------------------------------------------------
+import json
+import os, subprocess, sys
+import requests
+
+urlRoot = os.environ.get('catalogServiceUrl') or 'http://localhost:8080'
+urlDatasets = f'{urlRoot}/api/v1/datasets'
+urlDataRepoList = "https://jade.datarepo-dev.broadinstitute.org/api/repository/v1/snapshots"
+user = os.environ.get('gcloudUser') or 'datacatalogadmin@test.firecloud.org'
```
pshapiro4broad, line 26: Does python map upper case to camel case environment vars for you? I would expect the environment variables to be `GCLOUD_USER`. Also the instructions should mention which env vars can/should be used.

pshapiro4broad, line 26: Does python map upper case to camel case environment vars for you? I would expect the environment variables to be `GCLOUD_USER`. Also the instructions should mention which env vars can/should be used.

scripts/catalog-tests/generate-tdr-snapshot-metadata.py
```diff
@@ -0,0 +1,127 @@
...
+    print('---------------------------------------------------')
+    print(f'There was a problem updating dataset with metadata for ({datasetId}, {datasetTitle})')
+    print(json.loads(response.text)['message'])
+    print('\nRequest:')
+    print(f'\t{response.request.url}')
+    print(f'\t{response.request.headers}')
+    print(f'\t{response.request.body}')
+    print('---------------------------------------------------')
+  else:
+    print('success!', response.text)
```
pshapiro4broad, line 84: I know this is a super short script but could you refactor the common code here?

pshapiro4broad, line 84: I know this is a super short script but could you refactor the common code here?

scripts/catalog-tests/generate-tdr-snapshot-metadata.py
```diff
@@ -0,0 +1,127 @@
...
+    'Content-Type': 'application/json'
+  }
+  metadata = {
+      'storageSystem': 'tdr',
+      'storageSourceId': datasetId,
+      'catalogEntry': json.dumps(metadata)
+    }
+  response = requests.post(urlDatasets, headers=headers, data=json.dumps(metadata))
+
+  if response.status_code == 500:
```
pshapiro4broad, line 100: Maybe check for a non-2xx code instead? 500 is only one of many possible failure results.

pshapiro4broad, line 100: Maybe check for a non-2xx code instead? 500 is only one of many possible failure results.

pshapiro4broad (approved):
Looks OK overall, I have a few minor comments.

scripts/catalog-tests/generate-tdr-snapshot-metadata.py
lu-c: Good idea!

scripts/catalog-tests/generate-tdr-snapshot-metadata.py
lu-c: Updated this to use the caps'ed snake case!

scripts/catalog-tests/resources.json
```diff
@@ -0,0 +1,314 @@
...
+            {
+                "email": null,
+                "laboratory": "Molecular Immunity Unit, Department of Medicine",
+                "contactName": "Zewen,,Kelvin Tuong",
+                "institution": "University of Cambridge",
+                "projectRole": "experimental scientist",
+                "correspondingContributor": false
+            },
+            {
+                "email": "wteh@ebi.ac.uk",
```
fboulnois, line 81: Even though this info is public, I wonder if we should replace these with dummy emails, that way GitHub bots won't use this for spam.

fboulnois (approved):
One minor suggestion:

scripts/catalog-tests/resources.json
lu-c: That's a good point, I've updated the emails and the names to be randomly generated

PR #51, opened 2022-04-05
[DC-293] Add client id and secret to swagger UI to support google authentication

slucasbroad:
Add Client ID and Secret to use  Google Authentication in Swagger-UI.

Currently  `window.onload` fails to load with root cause unknown.


service/src/main/resources/templates/index.html
```diff
@@ -0,0 +1,156 @@
...
+      right: 0;
+      left: 0;
+      z-index: 100;
+    }
+  </style>
+</head>
+
+<body>
+<div id="swagger-ui"></div>
+<div style="background-color: #cfc ; padding: 10px; border: 1px solid green;">Sending but swagger not deplaying box</div>
```
fboulnois, line 75:
```suggestion
```

service/src/main/resources/templates/index.html
```diff
@@ -0,0 +1,156 @@
...
+<script src="/webjars/swagger-ui-dist/swagger-ui-bundle.js"> </script>
+<script src="/webjars/swagger-ui-dist/swagger-ui-standalone-preset.js"> </script>
+
```
fboulnois, line 78:
```suggestion
```

service/src/main/resources/templates/index.html
```diff
@@ -0,0 +1,156 @@
...
+    // End Swagger UI call region
+
+    ui.initOAuth({
+      clientId: [[${clientId}]],
+      scopes: "openid email profile"
+    })
+
+    window.ui = ui
+  }
+  </script>
```
fboulnois, line 154:
```suggestion
</script>
```

render_configs.sh
```diff
@@ -3,16 +3,19 @@ VAULT_TOKEN=${2:-$(cat "$HOME"/.vault-token)}
 
 VAULT_ADDR="https://clotho.broadinstitute.org:8200"
 COMMON_VAULT_PATH="secret/dsde/terra/kernel/$ENV/common"
+ECM_VAULT_PATH="secret/dsde/terra/kernel/$ENV/$ENV/catalog"
```
fboulnois, line 6:
```suggestion
CATALOG_VAULT_PATH="secret/dsde/terra/kernel/$ENV/$ENV/catalog"
```

render_configs.sh
```diff
@@ -3,16 +3,19 @@ VAULT_TOKEN=${2:-$(cat "$HOME"/.vault-token)}
...
 
+SERVICE_OUTPUT_LOCATION="$(dirname "$0")/service/src/main/resources/rendered"
 INTEGRATION_OUTPUT_LOCATION="$(dirname "$0")/integration/src/main/resources/rendered"
 
 if ! [ -x "$(command -v vault)" ]; then
   VAULT_COMMAND="docker run --rm -e VAULT_TOKEN=$VAULT_TOKEN -e VAULT_ADDR=$VAULT_ADDR vault:1.7.3 $VAULT_COMMAND"
 fi
 
 $VAULT_COMMAND -field=data -format=json "secret/dsde/firecloud/$ENV/common/firecloud-account.json" >"$INTEGRATION_OUTPUT_LOCATION/user-delegated-sa.json"
+$VAULT_COMMAND -field=swagger-client-id "$ECM_VAULT_PATH" >"$SERVICE_OUTPUT_LOCATION/swagger-client-id"
```
fboulnois, line 18:
```suggestion
$VAULT_COMMAND -field=swagger-client-id "$CATALOG_VAULT_PATH" >"$SERVICE_OUTPUT_LOCATION/swagger-client-id"
```

fboulnois (commented):
A few suggestions:

service/src/main/java/bio/terra/catalog/api/PublicApiController.java
```diff
@@ -27,13 +30,19 @@ public PublicApiController(
...
+          new BufferedReader(
+              new InputStreamReader(
+                  new ClassPathResource("rendered/swagger-client-id").getInputStream(),
+                  StandardCharsets.UTF_8))) {
+        clientId = reader.readLine();
+      }
     } catch (IOException e) {
       log.error(
           "It doesn't look like configs have been rendered! Unable to parse swagger client id.", e);
+      clientId = "";
```
fboulnois, line 45: Is this line still necessary?

service/src/main/java/bio/terra/catalog/api/PublicApiController.java
```diff
@@ -44,4 +69,10 @@ public ResponseEntity<VersionProperties> getVersion() {
   public String index() {
     return "redirect:swagger-ui.html";
   }
+
+  @RequestMapping(value = "/swagger-ui.html")
```
fboulnois, line 73:
```suggestion
  @RequestMapping(value = "/swagger-ui.html", method = RequestMethod.GET)
```

I think this will resolve the SonarQube complaint.

service/src/main/resources/templates/index.html
```diff
@@ -0,0 +1,154 @@
...
+      plugins: [
+        SwaggerUIBundle.plugins.DownloadUrl,
+        pinLoginPlugin,
+        clearValidator
+      ],
+      layout: 'StandaloneLayout',
+      displayOperationId: true,
+      docExpansion: 'none',
+      defaultModelsExpandDepth: -1, // hide the huge list of schemas under the routes
+      defaultModelExpandDepth: 2, // affects the schema shown for a request or response
```
pshapiro4broad, line 132: This HTML has changes which may be useful for TDR but aren't necessary for catalog. In the case of catalog, there's no need to hide the list of schemas since they're not very large (also it's collapsed by default).

The HTML is also different from the default as it loads the page with the API groups (`datasets` and `public`) collapsed. This is fine for TDR, which has many API groups and many APIs, but isn't necessary for catalog as it doesn't yet have more than a page of APIs. Can you revert the TDR specific changes, or fix the code so that it loads the way it did before?

service/src/main/resources/templates/index.html
```diff
@@ -0,0 +1,154 @@
...
+      oauth2RedirectUrl: window.location.protocol + '//' + window.location.host + '/webjars/swagger-ui-dist/oauth2-redirect.html',
+      tagsSorter: function (a, b) {
```
pshapiro4broad, line 134: Is this needed? Our tags are all lowercase so I don't think we need a custom tag sorter method.

service/src/main/resources/templates/index.html
```diff
@@ -0,0 +1,154 @@
...
+        } else {
+          return a < b ? -1 : 1;
+        }
+      }
+    })
+    // End Swagger UI call region
+
+    ui.initOAuth({
+      clientId: [[${clientId}]],
+      scopes: "openid email profile"
```
pshapiro4broad, line 147: I wonder if there's a way to get the scopes from the openapi yaml instead of hardcoding them here?

service/src/main/java/bio/terra/catalog/api/PublicApiController.java
```diff
@@ -4,22 +4,47 @@
...
     this.statusService = statusService;
     this.versionConfiguration = versionConfiguration;
+
+    String clientId = "";
+
+    try {
+      try (var reader =
+          new BufferedReader(
+              new InputStreamReader(
+                  new ClassPathResource("rendered/swagger-client-id").getInputStream(),
```
pshapiro4broad, line 39: For what it's worth, you don't need to use the Spring API `ClassPathResource()` here. You can instead use the Java API `getResourceAsStream()`: `getClass().getResourceAsStream("/rendered/swagger-client-id")`

This also lets us render the value into the `build` directory instead of the source directory. Although doing it that way means that if a user happens to clean their build tree they'll need to re-render the config, which is probably the reason why ECM has it set to work this way.

slucasbroad: code was suggested  to change it to 
`try { clientId =
          new String(
              new ClassPathResource("rendered/swagger-client-id").getInputStream().readAllBytes(),
              StandardCharsets.UTF_8);
    } `
should I change it to your way?

service/src/main/resources/templates/index.html
slucasbroad: from what I understand from html and and the`/webjars/swagger-ui-dist/swagger-ui.css`, I think the page is coded to look for the clientId value and I would need to mess with the webjar to put it into scopes I think.

service/src/main/java/bio/terra/catalog/api/PublicApiController.java
```diff
@@ -44,4 +64,10 @@ public ResponseEntity<VersionProperties> getVersion() {
   public String index() {
     return "redirect:swagger-ui.html";
   }
+
+  @GetMapping(value = "/swagger-ui.html")
+  public String getSwagger(Model model) {
+    model.addAttribute("clientId", swaggerClientId);
+    return "index";
+  }
```
fboulnois, line 72: SonarQube is complaining about the test coverage for these two methods. It should be relatively easy to unit test these methods.

slucasbroad: how do I set up a test?

pshapiro4broad:
This PR looks good to me. I think it's OK to ignore the Sonar check and use the red checkbox to override that result. The % uncovered is too low (the quality gate is 80%) but the total number of uncovered lines is small (2) and I don't think it's worth getting coverage on these lines.

pshapiro4broad (approved):

PR #53, opened 2022-04-06
DC-274: integration tests with snapshots

pshapiro4broad:
Add an integration test for all service operations. The test creates, updates, retrieves and deletes a catalog entry for a TDR snapshot.

While testing we found that the HTTP status code for a missing catalog entry record was 500, but should've been 404. This PR also fixes this behavior and adds a unit test for this case.

Note that permissions tests are not part of this PR, that work will be done in https://broadworkbench.atlassian.net/browse/DC-297


integration/src/main/resources/servers/catalog-local.json
```diff
@@ -3,6 +3,7 @@
   "description": "Catalog Local env.",
 
   "catalogUri": "http://localhost:8080",
+  "datarepoUri": "https://jade.datarepo-dev.broadinstitute.org/",
```
fboulnois, line 6: Do we need this to change depending on which environment we're running in?

integration/build.gradle
```diff
@@ -16,6 +16,7 @@ dependencies {
     implementation 'org.slf4j:slf4j-api'
     implementation 'org.glassfish.jersey.inject:jersey-hk2'
     implementation 'org.junit.jupiter:junit-jupiter-api'
+    implementation 'org.hamcrest:hamcrest'
```
fboulnois, line 19: Is there a strong reason to switch to hamcrest for testing? Easier use of assertions? I see without it the assertions are a few more lines of code, but the perhaps the tradeoff of adding a new package is worth it?

integration/src/main/java/scripts/client/DataRepoClient.java
```diff
@@ -0,0 +1,49 @@
...
+import com.google.auth.oauth2.GoogleCredentials;
+import java.io.IOException;
+import java.util.List;
+import java.util.Objects;
+
+public class DataRepoClient extends ApiClient {
+
+  // Required scopes for client tests include the usual login scopes and GCP scope.
+  public static final List<String> TEST_USER_SCOPES =
+      List.of("openid", "email", "profile", "https://www.googleapis.com/auth/cloud-platform");
```
fboulnois, line 16: Should we make this configurable generally and select them for the test? It may be useful for our swagger page too.

service/src/main/java/bio/terra/catalog/api/GlobalExceptionHandler.java
```diff
@@ -1,38 +1,16 @@
...
-
-  private ErrorReport buildErrorReport(@Nonnull Throwable e) {
-    StringBuilder combinedCauseString = new StringBuilder();
-    for (Throwable cause = e; cause != null; cause = cause.getCause()) {
-      combinedCauseString.append("cause: ").append(cause).append(", ");
-    }
-    logger.error("Global exception handler: " + combinedCauseString, e);
-    return new ErrorReport().message(e.getMessage());
+  @Override
+  public ErrorReport generateErrorReport(Throwable ex, HttpStatus statusCode, List<String> causes) {
```
fboulnois, line 13:
```suggestion
  public ErrorReport reportFromError(Throwable ex, HttpStatus statusCode, List<String> causes) {
```

integration/src/main/java/scripts/testscripts/SnapshotDatasetOperations.java
```diff
@@ -0,0 +1,165 @@
...
+    assertThat(client.getStatusCode(), is(HttpStatusCodes.STATUS_CODE_NO_CONTENT));
+
+    // Verify delete success.
+    var apiException = assertThrows(ApiException.class, () -> datasetsApi.getDataset(datasetId));
+    assertThat(apiException.getCode(), is(HttpStatusCodes.STATUS_CODE_NOT_FOUND));
+    assertThat(apiException.getResponseBody(), containsString(datasetId.toString()));
+    datasetId = null;
+  }
+
+  private void validateListResult(List<Object> listResult) {
```
fboulnois, line 138:
```suggestion
  private void datasetsHaveEntriesAndRoles(List<Object> datasetList) {
```

fboulnois (approved):
Overall looks good, a few suggestions. I also watched an interesting video on clean code so I'm trying to put their suggestions to practice :)

integration/src/main/resources/servers/catalog-local.json
pshapiro4broad: `local` works for both local development and `dev` tests, since the dev run is "local" to the Github runner it runs on.

I think we do need additional server configs for the various test environments we need to support. I can make another ticket for that work, and the work to find out how to have QA run them.

integration/build.gradle
pshapiro4broad: In most cases, hamcrest produces much better messages about failures; specifically it's better at showing differences between expected and actual data. It can be a bit wordier and does have its own language so there are some downsides too. Also I don't think there's a good equivalent of `assertThrows()` built in.

So the main differences are matters of personal preference.

integration/src/main/java/scripts/client/DataRepoClient.java
pshapiro4broad: I agree, this should be consolidated and not duplicated across different places. I'll look at cleaning that up, if it's too much work for now I'll make a ticket for it.

integration/src/main/java/scripts/testscripts/SnapshotDatasetOperations.java
pshapiro4broad: This is asserting that the result of the list operation contains a specific dataset with specific attributes. So maybe `resultHasDatasetWithRoles()`? I guess in general I think about method names as being <action><thing> so "validate" is the action and "list result" is the thing. Also I think about a `has` method as usually returning a `boolean`, which this doesn't do.

service/src/main/java/bio/terra/catalog/api/GlobalExceptionHandler.java
pshapiro4broad: I like this idea but it's an `Override` of a tcl method so we don't have control over the name.

integration/src/main/java/scripts/testscripts/SnapshotDatasetOperations.java
fboulnois: That sounds reasonable 👍

service/src/main/java/bio/terra/catalog/api/GlobalExceptionHandler.java
fboulnois: Oh yeah whoops :(

integration/src/main/resources/servers/catalog-local.json
pshapiro4broad: Created https://broadworkbench.atlassian.net/browse/DC-299

integration/src/main/java/scripts/testscripts/GetVersion.java
```diff
@@ -1,36 +1,30 @@
...
+    CatalogClient client = new CatalogClient(server);
+    var publicApi = new PublicApi(client);
 
     var versionProperties = publicApi.getVersion();
 
     // check the response code
-    var httpCode = publicApi.getApiClient().getStatusCode();
-    assertEquals(HttpStatusCodes.STATUS_CODE_OK, httpCode);
-    log.info("Service status return code: {}", httpCode);
+    assertThat(publicApi.getApiClient().getStatusCode(), is(HttpStatusCodes.STATUS_CODE_OK));
```
s-rubenstein, line 22: Could this be `client.getStatusCode()`, as above, for consistency?

integration/src/main/java/scripts/testscripts/GetStatus.java
```diff
@@ -1,26 +1,20 @@
...
 
 public class GetStatus extends TestScript {
-
-  private static final Logger log = LoggerFactory.getLogger(GetStatus.class);
-
   @Override
   public void userJourney(TestUserSpecification testUser) throws Exception {
-    log.info("Checking the service status endpoint.");
-    var publicApi = new PublicApi(new CatalogClient(server));
+    var client = new CatalogClient(server);
```
s-rubenstein, line 15: Why move this to an intermediate variable, as opposed to accessing through `publicApi` as before? 

I think it could go either way, just curious about rationale.

integration/src/main/java/scripts/testscripts/SnapshotDatasetOperations.java
```diff
@@ -0,0 +1,165 @@
...
+            .description("catalog test snapshot")
+            .profileId(dataset.getDefaultProfileId())
+            .addContentsItem(
+                new SnapshotRequestContentsModel()
+                    .datasetName(TEST_DATASET_NAME)
+                    .mode(SnapshotRequestContentsModel.ModeEnum.BYFULLVIEW));
+    var createJobId = snapshotsApi.createSnapshot(request).getId();
+
+    var jobApi = new JobsApi(datarepoClient);
+    JobModel job;
```
s-rubenstein, line 74: Should this explicitly have the `jobStatus` set to RUNNING? Right now that is the default, but just curious if that could change.

integration/src/main/java/scripts/client/DataRepoClient.java
pshapiro4broad: I've updated this to use a constant from terra-common-lib.

integration/src/main/java/scripts/testscripts/SnapshotDatasetOperations.java
pshapiro4broad: The code doesn't look at `job` until after `jobApi.retrieveJob(createJobId)` is called. So the contents of `job` at this point (which is `null`) doesn't matter.

it's true that if the Job API adds a new state, like `IN_QUEUE` or `PENDING` then this code could break. In that case maybe checking for `... == SUCCEEDED || .. == FAILED` would make more sense. But that could break too if another end state like `ABORTED` is added. My thinking is that if this does change in the future, the test will fail and then we'll have to update the code to match the current API behavior, so I wouldn't worry too much about planning for change now.

integration/src/main/java/scripts/client/CatalogClient.java
```diff
@@ -39,7 +34,8 @@ public CatalogClient(ServerSpecification server, TestUserSpecification testUser)
 
     if (testUser != null) {
       GoogleCredentials userCredential =
-          AuthenticationUtils.getDelegatedUserCredential(testUser, TEST_USER_SCOPES);
+          AuthenticationUtils.getDelegatedUserCredential(
+              testUser, AuthenticationUtils.userLoginScopes);
```
lu-c, line 38: Where are these userLoginScopes set from? (just learning question)

integration/src/main/java/scripts/testscripts/SnapshotDatasetOperations.java
```diff
@@ -0,0 +1,165 @@
...
+    assertThat(client.getStatusCode(), is(HttpStatusCodes.STATUS_CODE_OK));
+    assertThat(datasetId, notNullValue());
+    log.info("created dataset " + datasetId);
+
+    // Retrieve the entry
+    assertThat(datasetsApi.getDataset(datasetId), is(METADATA_1));
+    assertThat(client.getStatusCode(), is(HttpStatusCodes.STATUS_CODE_OK));
+
+    // Retrieve all datasets
+    var datasets = datasetsApi.listDatasets();
```
lu-c, line 115: Are these methods actually making a new rest call? It seems a bit like they're making internal calls to the services?

lu-c (approved):
This is great! We could use some cases that test failures, but there's no reason they cant be added later in th e future :)

integration/src/main/java/scripts/testscripts/SnapshotDatasetOperations.java
pshapiro4broad: Yes, this is making an API call to a server that's running in a separate process. The `ApiClient` class is the swagger generated client code which is a wrapper around `javax.ws.rs.client.Client`. One way to see this in action is to run the tests locally, where you can see the HTTP requests in the locally running server logs.

integration/src/main/java/scripts/client/CatalogClient.java
pshapiro4broad: `AuthenticationUtils` is from the terra testrunner library, and contains some code commonly needed when writing tests. `AuthenticationUtils.userLoginScopes` is a hardcoded list of OAuth scopes that's typically used when a user logs in. This API "logs in" `testUser` using these scopes and returns a credential, which is then used to get an access (aka Bearer) token. Once we have a bearer token, we can make authenticated API requests as `testUser`.

These scopes are the same as specified in our openapi yaml: https://github.com/DataBiosphere/terra-data-catalog/blob/main/common/openapi.yml#L281

pshapiro4broad:
> This is great! We could use some cases that test failures, but there's no reason they cant be added later in th e future :)

I agree, although I think the next most important set of tests to write are ones involving permissions. I created https://broadworkbench.atlassian.net/browse/DC-297 for this work.

PR #55, opened 2022-04-08
Add spotbugs html reports

fboulnois:


buildSrc/src/main/groovy/bio.terra.catalog.java-common-conventions.gradle
```diff
@@ -83,8 +83,10 @@ spotbugs {
 }
 spotbugsMain {
     reports {
-        sarif {
-            enabled = true
+        if (isCiServer) {
+            sarif.enabled = true
+        } else {
+            html.enabled = true
```
pshapiro4broad, line 89: There's probably no downside to always emit both forms in both cases. 
```suggestion
        sarif.enabled = true
        html.enabled = true
```

Also I've been thinking that we can spotbugs completely as I believe sonarqube includes spotbugs as part of its static analysis. But I haven't verified this theory yet...

pshapiro4broad (approved):
Looks OK to me, although I think enabling both might be best as someone might want `sarif` locally too, and there's no great cost in emitting both (AFAIK).

lu-c (approved):
Um, wow, short pr, looks good to me @.@

buildSrc/src/main/groovy/bio.terra.catalog.java-common-conventions.gradle
fboulnois: Unfortunately you can't generate `html` and `sarif` reports simultaneously.

PR #57, opened 2022-04-11
DC-308: Remove Temurin from brewfile and docs

fboulnois:
Homebrew has updated to Temurin 18, which is incompatible with our project. Remove the installation instructions in our project.


docs/getting-started.md
```diff
@@ -106,8 +106,6 @@ and **Community** (open-source). We recommend the Ultimate Edition to Broad
 employees for its database navigation capabilities. Alternatively, the Community
 Edition has all the features needed for development, and this version can be
 installed by switching `intellij-idea` with `intellij-idea-ce` in the Brewfile.
-8. [Temurin](https://adoptium.net/) is the code name for the free and
```
pshapiro4broad, line 109: Maybe replace this with a step elsewhere to install JDK 17 LTS from https://adoptium.net ?

pshapiro4broad (approved):

docs/getting-started.md
fboulnois: I think the three choices are (in ranked order of preference):

1. Include the alternate brew casks (the question is, does this work with the brewfile)
2. Let IntelliJ manage the Java 17 installation
4. Install the JDK17 from adoptium

pshapiro4broad: There's a another option which is to let `gradle` download the JDK. I haven't used this myself but it sounds like it would work. https://docs.gradle.org/current/userguide/toolchains.html#sec:provisioning. Once gradle has downloaded the JDK IntelliJ can be configured to use it.

pshapiro4broad: If we don't have an immediate answer maybe the best thing for now is to remove this wording and add a backlog task to look into this and then update the README after investigating.

s-rubenstein: I would probably lean towards options 2 or 3, because it gives us the most control. 1 has the chance of running into the same problem down the line.

fboulnois: @s-rubenstein that's a good point. Do you think we should defer making a decision to a future ticket or should I add a section with option 2 or 3?

s-rubenstein: I would add a section here, just so that the getting started docs continue to have a comprehensive list

fboulnois: I added a section for Java 17.

s-rubenstein (approved):
LGTM

PR #58, opened 2022-04-12
DC-268 create admin group and policies

tcjiang9:


scripts/bootstrapCatalogWithSam.sh
```diff
@@ -0,0 +1,34 @@
...
+AUTH_TOKEN=$2
+# resource_id such as broad-tdc-dev, broad-tdc-perf etc.
+RESOURCE_ID=$3
+ADMIN_GROUP_NAME=$4
+YOUR_EMAIL=$5
+
+# Create the admin group for the env
+curl -X POST -H 'Content-Length: 0' -H "Authorization: Bearer ${AUTH_TOKEN}" "${SAM_URL}/api/groups/v1/${ADMIN_GROUP_NAME}"
+# Add the user as an admin of the ops group
+curl -X PUT -H 'Content-Length: 0' -H "Authorization: Bearer ${AUTH_TOKEN}" "${SAM_URL}/api/groups/v1/${ADMIN_GROUP_NAME}/admin/${YOUR_EMAIL}"
```
pshapiro4broad, line 21: These are using `-H 'Content-Length: 0'` but I don't think that's required. Do they work without this directive?

scripts/bootstrapCatalogWithSam.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
```
pshapiro4broad, line 1: I think the name "bootstrap" is confusing here, also I think the one bash script we have now uses `_` instead of camel case so this filename should follow that convention. Also "catalog" seems redundant since we aren't going to have other sam initialization scripts in this repo. Maybe `create_groups_in_sam.sh`?

pshapiro4broad (changes requested):
I think we can improve upon the script from TDR in a few ways, to make it easier / less error-prone to run.

There are two other parts to this change. One is to update the code to use the new resource ID; currently the code is using `catalog` for all environments. If we use a different name for each environment, it will have to be configured using `application.yml` and possibly passed through using our helm chart. The other part is to add all data explorer team members to these groups in all environments. If these aren't done as part of this work, please create tickets for them in JIRA (if the tickets don't already exist).

scripts/bootstrapCatalogWithSam.sh
s-rubenstein, line 1: Potentially stupid question: Why write this script in bash? It seems like we have a lot of intermediate variables being stored in env, which can both leave data around and override local env variables. If we use a scripting language like python, we get more power to use things like local variables and arg parsing libraries.

We could also potentially get some safeguards using the `gcloud-firecloud` library https://github.com/broadinstitute/firecloud-develop/tree/dev/run-context/live/scripts/migrations/gcloud_firecloud (although I don't know if it was something that has been opened up to general use yet, CI was working on it and were discussing that it could be useful for a lot of teams' scripting)

scripts/bootstrapCatalogWithSam.sh
```diff
@@ -0,0 +1,34 @@
...
+
+if [[ $# -lt 5 ]]; then
+    echo "usage $0 <sam-url> <auth-token> <resource-id> <admin-group-name> <your-email>"
+    exit 1
+fi
+
+RESOURCE_TYPE="catalog"
+SAM_URL=$1
+AUTH_TOKEN=$2
+# resource_id such as broad-tdc-dev, broad-tdc-perf etc.
```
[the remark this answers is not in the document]
s-rubenstein: I like that as well, simpler and more aligned with the repo.

scripts/bootstrapCatalogWithSam.sh
```diff
@@ -0,0 +1,34 @@
...
+RESOURCE_ID=$3
+ADMIN_GROUP_NAME=$4
+YOUR_EMAIL=$5
```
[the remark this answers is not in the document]
s-rubenstein: What is the intended email, also. Is it the user you use in the environment? Your {env.}firecloud.org account? Something else?

scripts/bootstrapCatalogWithSam.sh
```diff
@@ -0,0 +1,34 @@
...
+# Create the admin group for the env
+curl -X POST -H 'Content-Length: 0' -H "Authorization: Bearer ${AUTH_TOKEN}" "${SAM_URL}/api/groups/v1/${ADMIN_GROUP_NAME}"
+# Add the user as an admin of the ops group
+curl -X PUT -H 'Content-Length: 0' -H "Authorization: Bearer ${AUTH_TOKEN}" "${SAM_URL}/api/groups/v1/${ADMIN_GROUP_NAME}/admin/${YOUR_EMAIL}"
+
+# create the resource id
+curl -X POST -H 'Content-Length: 0' -H "Authorization: Bearer ${AUTH_TOKEN}" "${SAM_URL}/api/resources/v2/${RESOURCE_TYPE}/${RESOURCE_ID}"
+
+# Get the email address for the ops group
+GROUP_EMAIL=$(curl --header 'Content-Type: application/json' --header 'Accept: application/json' --header "Authorization: Bearer ${AUTH_TOKEN}" "${SAM_URL}/api/groups/v1/${ADMIN_GROUP_NAME}")
```
s-rubenstein, line 27: We use `-H` elsewhere for headers, could we do that here as well for consistency?

scripts/bootstrapCatalogWithSam.sh
tcjiang9: this just exists as something reviewable and a record for what was done, I only need to run it one time and we're done with it. I followed the pattern in TDR
https://github.com/DataBiosphere/jade-data-repo/blob/develop/datarepoBootstrap.sh

scripts/bootstrapCatalogWithSam.sh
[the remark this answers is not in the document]
tcjiang9: > The other part is to add all data explorer team members to these groups in all environments.
yeah I mentioned in Slack, I'll do this manually through the API

s-rubenstein (approved):
Pending Phil's changes, this looks good.

fboulnois (changes requested):
Before merging, can we run `shellcheck` on this script?

fboulnois (commented):
Ditto Sky's comment.

scripts/bootstrapCatalogWithSam.sh
fboulnois: I agree.

.github/workflows/build-and-test.yml
```diff
@@ -28,12 +28,15 @@ jobs:
         with:
           sarif_file: service/build/reports/spotbugs/main.sarif
 
-  black:
+  linters:
```
pshapiro4broad, line 31: Putting this scan in a GitHub action that's not a gradle task makes it harder to run locally. If it were a task, we could add it to `test` so it's run as part of `gradle build`. I don't know how hard this would be to do, so that can wait for a future PR if it's not easy to do now.

render_configs.sh
```diff
@@ -14,6 +16,6 @@ fi
 
 $VAULT_COMMAND -field=data -format=json "secret/dsde/firecloud/$ENV/common/firecloud-account.json" >"$INTEGRATION_OUTPUT_LOCATION/user-delegated-sa.json"
 
-if [ $ENV == perf ]; then
+if [ "$ENV" = "perf" ]; then
```
pshapiro4broad, line 19: I think the fix here is to quote ENV only. I'm pretty sure quoting perf has no effect.

pshapiro4broad (approved):
Looks good, just a few last comments. It's a good change to have the yaml use a consistent naming scheme.

.github/workflows/build-and-test.yml
pshapiro4broad: I checked and there's a shellcheck plugin, and spotless apparently contains black directly. I'm not sure how to configure spotless to use black on python code.

render_configs.sh
fboulnois: I think it's confusing to not quote perf -- it makes it look like a variable instead of a string. Many of the examples I can find online quote raw strings, including the two relevant shellcheck entries for this line: https://github.com/koalaman/shellcheck/wiki/SC3014 and https://github.com/koalaman/shellcheck/wiki/SC2086

service/src/main/java/bio/terra/catalog/iam/SamService.java
```diff
@@ -50,7 +50,7 @@ public boolean hasGlobalAction(AuthenticatedUserRequest userRequest, SamAction a
     ResourcesApi resourceApi = resourcesApi(accessToken);
     try {
       return SamRetry.retry(
-              () -> resourceApi.resourceActions(CATALOG_RESOURCE_TYPE, CATALOG_RESOURCE_ID))
```
fboulnois, line 53: It looks like SonarQube is complaining about a stray `CATALOG_RESOURCE_ID` somewhere in the code. Probably worth removing it if it isn't relevant.

fboulnois (approved):
lgtm but see suggestion above.

.github/workflows/build-and-test.yml
tcjiang9: @fboulnois I think this is something you added to my PR?

pshapiro4broad: I've created a separate PR for setting up gradle to do these scans.

render_configs.sh
pshapiro4broad: Probably a matter of individual taste; I'm used to not quoting but that's just me. The reason I avoid it is because bash doesn't have strings, and adding quotes to things makes it look like it does, which just furthers the confusion about bash and strings.

PR #59, opened 2022-04-13
DC-314: Fix nightly perf test notifications and errors

fboulnois:
The integration test script is missing the `datarepoUri`, and the notifications should always be sent.


pshapiro4broad (approved):
Whoops. Thanks for fixing this!

.github/workflows/nightly-tests.yml
```diff
@@ -84,6 +84,8 @@ jobs:
   notify-slack:
     needs: [ build, jib, test-runner ]
     runs-on: ubuntu-latest
+    if: always()
```
pshapiro4broad, line 87: WIth this change, is the `if always()` on line 103 below still required?

petesantos (approved):

PR #60, opened 2022-04-13
[DC-288] SL update catalog api to request no store for sensitive data returned

slucasbroad:
`getDataset()` and `listDatasets()` Responses are not stored to the user's cache. Add to the tests to check that the header is set to `noStore`.


fboulnois (approved):
Approved, there are other changes but they are not relevant to this PR.

pshapiro4broad (approved):
Looks good to me. After this merges and is deployed, we'll need to contact appsec to request another pentest scan so they can verify that this change addresses the issue.

pshapiro4broad:
The commit history is a bit messy but using Squash and Merge should hide all that so I wouldn't worry about it.

PR #63, opened 2022-04-14
Add preview metadata endpoint

s-rubenstein:
Tested via swagger.

I would love some eyes on this in draft mode, before I go ahead and write tests. There's a _lot_ of boilerplate and I'd love if folks had ideas for how to cut it down/if you all have a different approach. (I suppose I could just cast down the line, but casting always feels not great to me)

This code is functional, if not as clean as I would like it to be.


service/src/main/java/bio/terra/catalog/service/DatasetService.java
```diff
@@ -131,4 +140,90 @@ public DatasetId createDataset(
...
+    var snapshotMetadata = datarepoService.getPreviewMetadata(user, dataset.storageSourceId());
+    return new DatasetPreviewMetadataResponse()
+        .id(snapshotMetadata.getId())
+        .dataProject(snapshotMetadata.getDataProject())
+        .tables(convertDatarepoTablesToCatalogTables(snapshotMetadata.getTables()));
+  }
+
+  private List<TableModel> convertDatarepoTablesToCatalogTables(
+      List<bio.terra.datarepo.model.TableModel> datarepoTables) {
+    return Optional.ofNullable(datarepoTables)
```
pshapiro4broad, line 159: In general I don't think `Optional.ofNullable()` should be used as a replacement for a conditional. If you want to write this in a functional-ish way, you can use `Objects.requireNonNullElse()`. Personally I would use `!= null`.

Although, is it possible for `datrepoTables` to be `null` in practice? Ask in jade-core to be sure, but I don't think it is.

service/src/main/java/bio/terra/catalog/service/DatasetService.java
```diff
@@ -131,4 +140,90 @@ public DatasetId createDataset(
...
+                                .partitionMode(
+                                    convertDatarepoPartiotionMode(tableModel.getPartitionMode()))
+                                .datePartitionOptions(
+                                    convertDatarepoDatePartionsOptionsModel(
+                                        tableModel.getDatePartitionOptions()))
+                                .intPartitionOptions(
+                                    convertDatarepoIntPartionsOptionsModel(
+                                        tableModel.getIntPartitionOptions()))
+                                .rowCount(tableModel.getRowCount()))
+                    .collect(Collectors.toList()))
```
pshapiro4broad, line 179: In JDK 17 you can use `toList()`
```suggestion
                    .toList())
```

service/src/main/java/bio/terra/catalog/service/DatasetService.java
```diff
@@ -131,4 +140,90 @@ public DatasetId createDataset(
...
+        .orElse(Collections.emptyList());
+  }
+
+  private TableModel.PartitionModeEnum convertDatarepoPartiotionMode(
```
pshapiro4broad, line 183: typo
```suggestion
  private TableModel.PartitionModeEnum convertDatarepoPartitionMode(
```

common/openapi.yml
```diff
@@ -145,6 +145,24 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorReport'
+  /api/v1/datasets/{id}/preview-metadata:
```
fboulnois, line 148: Any thought between `/api/v1/datasets/{id}/preview-metadata` and `/api/v1/datasets/{id}/preview` ? Even though data repo calls this `metadata`, I think this is actually data (as opposed to the catalog entries, which I would consider metadata).

pshapiro4broad (commented):
Some comments. Thinking about this more, I think that one endpoint should only return the table names (that is, column name/types/etc aren't needed), and the other one would return the same as `/api/repository/v1/snapshots/{id}/data/{table}` does in current TDR.

common/openapi.yml
```diff
@@ -267,6 +286,115 @@ components:
...
+          type: integer
+          description: >
+            The largest value to partition within the target column. Any rows with a value larger than this will be unpartitioned.
+          format: int64
+        interval:
+          type: integer
+          description: >
+            The size to use when dividing the partitioning range into "buckets". (max - min) / (this value) cannot be larger than 4,000.
+          format: int64
+      description: Describes how an int partition should be configured.
```
fboulnois, line 396: I think something like this might be useful in the future and I think it's worth discussing. In the meantime I'm a fan of delivering the smallest unit of work that we can have good test coverage for, so I think we could leave this for a future ticket.

service/src/main/java/bio/terra/catalog/service/DatasetService.java
```diff
@@ -131,4 +140,90 @@ public DatasetId createDataset(
...
+                                    convertDataRepoColumnsToCatalogColumns(tableModel.getColumns()))
+                                .primaryKey(tableModel.getPrimaryKey())
+                                .partitionMode(
+                                    convertDatarepoPartiotionMode(tableModel.getPartitionMode()))
+                                .datePartitionOptions(
+                                    convertDatarepoDatePartionsOptionsModel(
+                                        tableModel.getDatePartitionOptions()))
+                                .intPartitionOptions(
+                                    convertDatarepoIntPartionsOptionsModel(
+                                        tableModel.getIntPartitionOptions()))
```
fboulnois, line 177: By removing some of the partitioning modes it should clean up this boilerplate a bit.

fboulnois (commented):
A few early comments:

common/openapi.yml
s-rubenstein: So, there's going to be a separate endpoint to `getPreviewTable`. I could call them `preview` and `preview-table`? But as a user, I wouldn't know what that meant.

README.md
```diff
@@ -99,3 +99,13 @@ generate a report, run using `--info`:
...
+## Common Errors:
+If you see the following error when running `./gradlew build`:
+```
+./gradlew build
+> Task :buildSrc:validatePlugins FAILED
+
+FAILURE: Build failed with an exception.
+```
+
+You should try installing gradle using the command `brew install gradle` for MacOS, and then run via `gradle build`
```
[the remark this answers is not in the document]
s-rubenstein: I think I am going to remove this section, because I am not confident enough that it wasn't related to some of the other things we ran into while setting up my env.

.gitignore
```diff
@@ -30,3 +30,6 @@ bin/
 
 # Mac directory metadata
 .DS_Store
+
+# PyEnv environment files
+.env/
```
fboulnois, line 35: Good addition, thank you!

fboulnois (approved):
This looks like it's in good shape!

common/openapi.yml
```diff
@@ -267,6 +286,59 @@ components:
...
+      type: object
+      properties:
+        tables:
+          type: array
+          items:
+            $ref: '#/components/schemas/TableModel'
+      description: >
+        DatasetPreviewTablesResponse returns information about the table structure of a dataset.
+
+    TableModel:
```
pshapiro4broad, line 299: I still have questions about this API and what it needs to provide. The API this is modeled after TDR's `/api/repository/v1/snapshots/{id}?include=TABLES`. However the UI doesn't need many of the fields that are in this response. I believe the only data needed by the UI is:
- the table names
- for each table, the number or rows (since we disable selecting empty tables in the UI)

Column names aren't needed as the column names are included in the preview API that returns the table contents.

I think it would be best to have the smallest API possible that satisfies the UI's needs, to make it easier to add support for different storage systems.

service/src/main/java/bio/terra/catalog/datarepo/DatarepoService.java
```diff
@@ -54,6 +56,19 @@ public Map<String, List<String>> getSnapshotIdsAndRoles(AuthenticatedUserRequest
...
   }
 
+  public SnapshotModel getPreviewTables(AuthenticatedUserRequest user, String snapshotId) {
+    try {
+      UUID id = UUID.fromString(snapshotId);
+      return snapshotsApi(user)
+          .retrieveSnapshot(
+              id,
+              List.of(
+                  SnapshotRetrieveIncludeModel.DATA_PROJECT, SnapshotRetrieveIncludeModel.TABLES));
```
pshapiro4broad, line 66: Is `SnapshotRetrieveIncludeModel.DATA_PROJECT` needed? I think we only use the `TABLES` content in the response.

service/src/main/java/bio/terra/catalog/service/DatasetService.java
```diff
@@ -131,4 +144,37 @@ public DatasetId createDataset(
     ensureActionPermission(user, dataset, SamAction.CREATE_METADATA);
     return datasetDao.create(dataset).id();
   }
+
+  public DatasetPreviewTablesResponse getDatasetPreviewTables(
+      AuthenticatedUserRequest user, DatasetId datasetId) {
+    var dataset = datasetDao.retrieve(datasetId);
+    ensureActionPermission(user, dataset, SamAction.READ_ANY_METADATA);
```
pshapiro4broad, line 151: Looking at this I'm realizing that the code in `DatarepoService.userHasAction()` isn't quite correct; it combines reader and discoverer operations and does the same check for both. So this check will pass even if a user only has discoverer permissions. Since TDR itself also does a permission check, and uses the current user's credentials, no data will be returned and the TDR API call will return an error (probably 403).

For now this is OK, but this makes me wonder if we should be doing a permission check here at all. For operations that only pass through to other systems, we're duplicating work done elsewhere and therefore have a chance to do it incorrectly (as we're doing now). For operations that modify the catalog database, this check is still necessary.

pshapiro4broad (commented):
A few more comments, mainly about what the UI code needs this API to provide.

service/src/main/java/bio/terra/catalog/datarepo/DatarepoService.java
s-rubenstein: Tables is the only thing that was needed, but for some reason `DATA_PROJECT` was also included in the UI. I am going to pull it.

service/src/main/java/bio/terra/catalog/service/DatasetService.java
pshapiro4broad: I created [DC-315](https://broadworkbench.atlassian.net/browse/DC-315) to track this issue.

service/src/main/java/bio/terra/catalog/datarepo/DatarepoService.java
```diff
@@ -54,6 +56,15 @@ public Map<String, List<String>> getSnapshotIdsAndRoles(AuthenticatedUserRequest
     }
   }
 
+  public SnapshotModel getPreviewTables(AuthenticatedUserRequest user, String snapshotId) {
+    try {
+      UUID id = UUID.fromString(snapshotId);
+      return snapshotsApi(user).retrieveSnapshot(id, List.of(SnapshotRetrieveIncludeModel.TABLES));
+    } catch (ApiException e) {
+      throw new DatarepoException(e.getMessage(), e.getCause());
```
pshapiro4broad, line 64: I forgot to comment on this before; in general I think it's best to include the caught exception when rethrowing. By rethrowing using `.getCause()` the stack crawl won't include the intermediate throw that resulted in `ApiException`. Is that intentional? I would expect to see instead
```suggestion
      throw new DatarepoException(e);
```

(Note that doing this will require adding another constructor to `DatarepoException`)

It also occurs to me that we'd want to make sure that this correctly maps the http result code on an access failure. I _think_ this can be done like:

```
throw new ErrorReportException(e, HttpStatus.resolve(e.getCode()));
```

pshapiro4broad (approved):
Looking at this again, I noticed something I missed before regarding error handling, sorry about that! Also it looks like you'll need a test for the controller code to satisfy sonarqube. Looks good otherwise.

service/src/main/java/bio/terra/catalog/datarepo/DatarepoException.java
```diff
@@ -1,7 +1,15 @@
...
-public class DatarepoException extends RuntimeException {
-  public DatarepoException(String message, Throwable cause) {
-    super(message, cause);
+import bio.terra.common.exception.ErrorReportException;
+import bio.terra.datarepo.client.ApiException;
+import org.springframework.http.HttpStatus;
+
+public class DatarepoException extends ErrorReportException {
+  public DatarepoException(String message, ApiException e) {
+    super(message, e.getCause(), null, HttpStatus.valueOf(e.getCode()));
```
pshapiro4broad, line 9: I'm pretty sure you want `HttpStatus.resolve()` not `valueOf()`, since `valueOf()` will throw if it can't map the status code, and we wouldn't want to throw in a constructor for an exception.
```suggestion
    super(message, e.getCause(), null, HttpStatus.resolve(e.getCode()));
```

Also here again I would use `e` not `e.getCause()`, is there a reason to not use `e`?
```suggestion
    super(message, e, null, HttpStatus.valueOf(e.getCode()));
```

s-rubenstein: No. I assumed typing-wise it wouldn't work and it does. Just some bad assumptions based off the existing code.

PR #64, opened 2022-04-14
DC-222: Add missing .py in README.md

fboulnois:
As I was debugging DC-222 I came across this issue.


pshapiro4broad (approved):

pshapiro4broad:
Is there a reason why we don't make our .py files executable and have users run them directly? If they have a venv set up then that should work OK.

slucasbroad (approved):

fboulnois:
> Is there a reason why we don't make our .py files executable and have users run them directly? If they have a venv set up then that should work OK.

I don't have a strong opinion about this.

PR #65, opened 2022-04-15
A script to set up a proxy to the catalog postgres database.

pshapiro4broad:
This script supports running a local proxy to connect to our cloud database. When run, it prints out a JDBC URL that can be pasted into IntelliJ or other SQL IDEs to view tables, etc. Note that this connects using the admin user so it can modify data, so please use it carefully when connecting to the prod database.

### DIFF
--- a/scripts/db_connect.sh
+++ b/scripts/db_connect.sh
@@ -3,7 +3,7 @@
 # This script requires google's cloud_sql_proxy on your PATH. One way to set this up:
 # $ brew cask install google-cloud-sdk
 # $ gcloud components install cloud_sql_proxy
-# $ PATH=$PATH:/usr/local/Caskroom/google-cloud-sdk/latest/google-cloud-sdk/bin/
+# $ PATH="$PATH:/usr/local/Caskroom/google-cloud-sdk/latest/google-cloud-sdk/bin"
 # OR
 # $ ln -s /usr/local/Caskroom/google-cloud-sdk/latest/google-cloud-sdk/bin/cloud_sql_proxy /usr/local/bin/cloud_sql_proxy
 #
--- a/scripts/db_connect.sh
+++ b/scripts/db_connect.sh
@@ -22,11 +22,16 @@ PORT=${PORT:-5431}
 VAULT_PATH="secret/dsde/terra/kernel/${ENV}/${ENV}/catalog/postgres"
 
 INSTANCE=$(vault read -field=data -format=json "${VAULT_PATH}/instance" |
-           jq -r '"\(.project):\(.region):\(.name)"')
+           jq -r '"\(.project):\(.region):\(.name)"')=tcp:$PORT
 
-JDBC_URL=jdbc:postgresql://localhost:$PORT/$(vault read -field=data -format=json "${VAULT_PATH}/db-creds" |
+DB_CREDS_DATA=$(vault read -field=data -format=json "${VAULT_PATH}/db-creds")
+
+JDBC_URL=jdbc:postgresql://localhost:$PORT/$(echo "${DB_CREDS_DATA}" |
           jq -r '"\(.db)?user=\(.username)&password=\(.password)"')
 
-echo "Starting a proxy for $ENV. Connection: $JDBC_URL"
+PSQL_COMMAND=$(echo "${DB_CREDS_DATA}" |
+          jq -r '"psql postgresql://\(.username):\(.password)@localhost/\(.db)\\?port="')$PORT
+
+echo "Starting a proxy for $ENV. Connect using: \"$JDBC_URL\" or run: \"$PSQL_COMMAND\""
 
 cloud_sql_proxy -instances="${INSTANCE}" -dir=/tmp
--- a/scripts/db_connect.sh
+++ b/scripts/db_connect.sh
@@ -1,0 +1,34 @@
+#!/bin/sh
+
+# This script requires google's cloud_sql_proxy on your PATH. One way to set this up:
+# $ brew cask install google-cloud-sdk
+# $ gcloud components install cloud_sql_proxy
+# $ PATH=$PATH:/usr/local/Caskroom/google-cloud-sdk/latest/google-cloud-sdk/bin/
+#
+# Use this script to connect to cloud Postgres for a specific Catalog instance.
+# For example, to connect to the dev catalog database, run:
+# $ ENV=dev ./db-connect.sh
+#
+# The proxy will continue to run until you quit it using ^C.
+#
+# The default port used is 5431, to avoid conflicting with a locally running postgres which
+# defaults to port 5432. Use the environment variable PORT to override this setting.
+
+: "${ENV:?}"
+PORT=${PORT:-5431}
+
+VAULT_PATH="secret/dsde/terra/kernel/${ENV}/${ENV}/catalog/postgres"
+
+INSTANCE_DATA=$(vault read -field=data -format=json "${VAULT_PATH}/instance")
+NAME=$(echo "${INSTANCE_DATA}" | jq -r '.name')
+PROJECT=$(echo "${INSTANCE_DATA}" | jq -r '.project')
+REGION=$(echo "${INSTANCE_DATA}" | jq -r '.region')
+
+DB_DATA=$(vault read -field=data -format=json "${VAULT_PATH}/db-creds")
+DB=$(echo "${DB_DATA}" | jq -r '.db')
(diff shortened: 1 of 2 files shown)


pshapiro4broad:
Changing to Draft until I can write up more docs and maybe clean up the `jq` a bit.

scripts/db_connect.sh
```diff
@@ -0,0 +1,32 @@
...
+: "${ENV:?}"
+PORT=${PORT:-5431}
+
+VAULT_PATH=secret/dsde/terra/kernel/${ENV}/${ENV}/catalog/postgres
+
+INSTANCE=$(vault read -field=data -format=json "${VAULT_PATH}/instance" |
+           jq -r '"\(.project):\(.region):\(.name)"')
+
+JDBC_URL=jdbc:postgresql://localhost:$PORT/$(vault read -field=data -format=json "${VAULT_PATH}/db-creds" |
+          jq -r '"\(.db)?user=\(.username)&password=\(.password)"')
```
fboulnois, line 28: I don't think we can be sure that a `jdbc` connection is set up appropriately in Intellij or another tool, especially since some devs are using the community edition which I believe does not support database connections. Instead I would output the url (without jdbc) and perhaps also automatically connect through `psql`. All team members should have `psql` installed as part of the Postgres.app setup in our getting started docs. This would allow devs to have a default starting point if desired, but allow them to use other tools also.

scripts/db_connect.sh
```diff
@@ -0,0 +1,32 @@
...
+#
+# The proxy will continue to run until you quit it using ^C.
+#
+# The default port used is 5431, to avoid conflicting with a locally running postgres which
+# defaults to port 5432. Use the environment variable PORT to override this setting.
+
+: "${ENV:?}"
+PORT=${PORT:-5431}
+
+VAULT_PATH=secret/dsde/terra/kernel/${ENV}/${ENV}/catalog/postgres
```
fboulnois, line 22:
```suggestion
VAULT_PATH="secret/dsde/terra/kernel/${ENV}/${ENV}/catalog/postgres"
```

I recommend quoting all strings for clarity, consistency with POSIX sh, and potential wildcard expansion. Has this script been checked with shellcheck?

scripts/hca/README.md
```diff
@@ -27,6 +27,9 @@ python3 create-hca-collection.py
 
 ```sh
 source .env/bin/activate
+export USER_EMAIL=datacatalogadmin@test.firecloud.org
+# OR
+export USER_EMAIL=monster.pshapiro.dev@gmail.com
```
fboulnois, line 32:
```suggestion
# OR export your personal dev admin email address
```

Perhaps something like this?

scripts/db_connect.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/sh
+
+# This script requires google's cloud_sql_proxy on your PATH. One way to set this up:
+# $ brew cask install google-cloud-sdk
+# $ gcloud components install cloud_sql_proxy
+# $ PATH=$PATH:/usr/local/Caskroom/google-cloud-sdk/latest/google-cloud-sdk/bin/
```
fboulnois, line 6:
```suggestion
# $ PATH="$PATH:/usr/local/Caskroom/google-cloud-sdk/latest/google-cloud-sdk/bin"
```

`PATH` should be quoted and paths should not include a terminating slash.

fboulnois (commented):
A few suggestions:

scripts/db_connect.sh
pshapiro4broad: This code formats a JDBC URL and prints it to stdout so a user can cut & paste this and use it in IntelliJ or any other tool that supports JDBC URLs. I could remove the `jdbc` prefix, but without it, what tool would be able to use the URL?

I'll add code that emits the `psql` command as well as the URL, so users can pick the one that matches what they need.

scripts/db_connect.sh
pshapiro4broad: I can add the quotes back, but they have no effect in this case. Shellcheck doesn't flag this as an issue.

scripts/db_connect.sh
pshapiro4broad: IMO quoting `PATH` seems odd and not necessary. But it also seems harmless, so why not. I guess some user might have a path element with spaces in it.

scripts/db_connect.sh
pshapiro4broad: Oh, I didn't realize the psql command line accepted the `postgresql` link as an argument. I agree, I can make the code emit this instead.

pshapiro4broad:
Here's what the output currently looks like: (with the actual database username and password removed)
```
Starting a proxy for dev. Connect using: "jdbc:postgresql://localhost:5433/catalog?user=user&password=password" or run: "psql postgresql://user:password@localhost/catalog\?port=5433"
```

scripts/db_connect.sh
```diff
@@ -0,0 +1,37 @@
...
+# This script requires google's cloud_sql_proxy on your PATH. One way to set this up:
+# $ brew cask install google-cloud-sdk
+# $ gcloud components install cloud_sql_proxy
+# $ PATH="$PATH:/usr/local/Caskroom/google-cloud-sdk/latest/google-cloud-sdk/bin"
+# OR
+# $ ln -s /usr/local/Caskroom/google-cloud-sdk/latest/google-cloud-sdk/bin/cloud_sql_proxy /usr/local/bin/cloud_sql_proxy
+#
+# Use this script to connect to cloud Postgres for a specific Catalog instance.
+# For example, to connect to the dev catalog database, run:
+# $ ENV=dev ./db-connect.sh
```
s-rubenstein, line 12: If you have an existing `ENV`, what does this do?

s-rubenstein (commented):
Is there a ticket for this work?

scripts/db_connect.sh
pshapiro4broad: If you set an environment variable before a command, then that env var is only set for that command (vs exporting it, which will make it visible to all commands). If the env var is already set, the value set on the command line takes precedence.

pshapiro4broad:
```sh
$ cat test.sh
echo $X
$ export X="hello"
$ ./test.sh
hello
$ X="goodbye" ./test.sh
goodbye
$
```

scripts/db_connect.sh
```diff
@@ -0,0 +1,37 @@
...
+INSTANCE=$(vault read -field=data -format=json "${VAULT_PATH}/instance" |
+           jq -r '"\(.project):\(.region):\(.name)"')=tcp:$PORT
+
+DB_CREDS_DATA=$(vault read -field=data -format=json "${VAULT_PATH}/db-creds")
+
+JDBC_URL=jdbc:postgresql://localhost:$PORT/$(echo "${DB_CREDS_DATA}" |
+          jq -r '"\(.db)?user=\(.username)&password=\(.password)"')
+
+PSQL_COMMAND=$(echo "${DB_CREDS_DATA}" |
+          jq -r '"psql postgresql://\(.username):\(.password)@localhost/\(.db)\\?port="')$PORT
```
fboulnois, line 33:
```suggestion
          jq -r '"psql postgresql://\(.username):\(.password)@localhost/\(.db)\?port="')$PORT
```

Is there an extra backslash here? I noticed it in your example output.

fboulnois (approved):
Overall looks reasonable, one question:

scripts/db_connect.sh
pshapiro4broad: Yeah, it's needed because `?` is a shell globbing character and it needs to be quoted to prevent your shell from trying to match the "postgresql..." string to a file in the current directory. Kind of ugly but it works.

scripts/db_connect.sh
s-rubenstein: Perfect. Although weirdly when I ran:
```
$ export foo="bar"   
$ foo="baz" echo $foo
bar
```
But I think that sys level commands might work differently

s-rubenstein (approved):
LGTM

scripts/db_connect.sh
pshapiro4broad: When you write `foo="baz" echo $foo`, the `$foo` substitution occurs before the anything in the command line is executed. This can be useful sometimes, e.g. `PATH=$PATH:/tmp/bin something`. If `echo` itself supported a way to print an environment variable then you could change `$foo` on the command line:
```sh
$ export X="hello"
$ echo "" | X="goodbye" awk '{print ENVIRON["X"]}'
goodbye
```

PR #67, opened 2022-04-15
[DC-298] Sl For response consistency and clarity, listDatasets should return an empty result set, rather than no result field at all, when there are no results

slucasbroad:
listDatasets(), return {result:[]}, as opposed to {}.

### DIFF
--- a/common/openapi.yml
+++ b/common/openapi.yml
@@ -236,6 +236,7 @@ components:
 
     DatasetsListResponse:
       type: object
+      required: [result]
       properties:
         result:
           type: array
--- a/scripts/catalog-tests/.env/lib/python3.9/site-packages/pip/_vendor/packaging/__about__.py
+++ b/scripts/catalog-tests/.env/lib/python3.9/site-packages/pip/_vendor/packaging/__about__.py
@@ -1,0 +1,26 @@
+# This file is dual licensed under the terms of the Apache License, Version
+# 2.0, and the BSD License. See the LICENSE file in the root of this repository
+# for complete details.
+
+__all__ = [
+    "__title__",
+    "__summary__",
+    "__uri__",
+    "__version__",
+    "__author__",
+    "__email__",
+    "__license__",
+    "__copyright__",
+]
+
+__title__ = "packaging"
+__summary__ = "Core utilities for Python packages"
+__uri__ = "https://github.com/pypa/packaging"
+
+__version__ = "21.0"
+
+__author__ = "Donald Stufft and individual contributors"
+__email__ = "donald@stufft.io"
+
+__license__ = "BSD-2-Clause or Apache-2.0"
+__copyright__ = "2014-2019 %s" % __author__


fboulnois (approved):

service/src/test/java/bio/terra/catalog/api/DatasetApiControllerTest.java
```diff
@@ -74,6 +74,20 @@ void listDatasets() throws Exception {
         .andExpect(jsonPath("$.result[0].id").value("id"));
   }
 
+  @Test
+  void EmptyListDatasets() throws Exception {
+    DatasetsListResponse response = new DatasetsListResponse();
+    ObjectNode node = new ObjectMapper().createObjectNode();
+    response.addResultItem(node);
+    when(datasetService.listDatasets(user)).thenReturn(response);
```
s-rubenstein, line 82: Because this datasetService call is mocked, I am not sure that this test adds much.

I would probably rather see a test that calls the underlying method (maybe with a mock on `datasetDAO` to ensure we get no values back from the database?) and at the end something like:
`assertThat(response.getResults().getSize()).isZero()`

lu-c (changes requested):
Approved, but want Sky's comment to be addressed :)

service/src/test/java/bio/terra/catalog/api/DatasetApiControllerTest.java
pshapiro4broad: I think this is the correct way to test this. The purpose of this test is to see if the JSON result is as expected or not. This does require that we create a response object that correctly matches what we'd get if there are no datasets to return, and I think removing the call to `response.addResultItem()` is required to do that.

Regarding `assertThat(response.getResults().getSize()).isZero()`, I think `.andExpect(jsonPath("$.result[0]").isEmpty())` fulfills this requirement (maybe your comment refers to an earlier version of this test?).

> I would probably rather see a test that calls the underlying method (maybe with a mock on datasetDAO to ensure we get no values back from the database?

If this test were to mock at the DAO layer, then it would have to use a "real" `DatasetService` instead of a mocked one, which would change the way all test methods in this class works. In general I think we want a test to mock everything except the class under test itself. (Except for DAO tests, which require a live database.)

pshapiro4broad (changes requested):

service/src/test/java/bio/terra/catalog/api/DatasetApiControllerTest.java
s-rubenstein: In my mind, the place to test is at the place where if it found no items, it generated the response object. My concern here is that we're mocking out the actual response object, which means we aren't testing our code, we're testing what swagger is doing. Given that the change was just done in swagger, I actually am not sure we should test this anymore.

If we do need to test it, I would probably do:
```
DatasetsListResponse response = new DatasetsListResponse();
assertThat(response.getResultItems().size(), assertDoesNotThrow())
```

Because that is actually the behavior that we want to check here (that doing operations on the result set doesn't fail even if it is empty)

pshapiro4broad (approved):
Looks good, although you still need to rename the test method `EmptyListDatasets()` to `emptyListDatasets()`

lu-c (approved):

service/src/test/java/bio/terra/catalog/api/DatasetApiControllerTest.java
```diff
@@ -74,6 +74,18 @@ void listDatasets() throws Exception {
...
+  @Test
+  void EmptyListDatasets() throws Exception {
+    DatasetsListResponse response = new DatasetsListResponse();
+    when(datasetService.listDatasets(user)).thenReturn(response);
+    mockMvc
+        .perform(get(API))
+        .andExpect(status().isOk())
+        .andExpect(header().string("Cache-Control", "no-store"))
+        .andExpect(content().contentType(MediaType.APPLICATION_JSON))
+        .andExpect(jsonPath("$.result").isEmpty());
```
pshapiro4broad, line 86: This check will produce a false positive. That is, even if you back out the yaml change, this test will continue to pass. I think that's because `isEmpty()` matches both missing and empty arrays in the JSON response. Using this assert instead will correctly test the change. You can verify this by reverting the yaml change locally and running the test.
```suggestion
        .andExpect(jsonPath("$.result").isArray());
```

pshapiro4broad (changes requested):
The assert isn't quite correct, try out my change and verify that it correctly tests the configuration change. Also don't forget to rename the test method to a lower case name.

common/openapi.yml
```diff
@@ -236,6 +236,7 @@ components:
 
     DatasetsListResponse:
       type: object
+      required: [result]
```
lu-c, line 239:
```suggestion
      required: [ result ]
```
suggestion: add spaces in the array, since that's the style the rest of the code is in

pshapiro4broad (approved):
thanks!

s-rubenstein (approved):
LGTM

PR #68, opened 2022-04-19
DC-300: Update the swagger-client-id path

fboulnois:
Previously, we had client ids which lived at the following paths:

1. `secret/dsde/terra/kernel/dev/dev/catalog/swagger-client-id`
2. `secret/dsde/terra/kernel/dev/dev/catalog`

To be consistent with other services, the client id should live at `{service}/swagger-client-id`, however we had an invalid client id at that path and a valid one at the `/catalog` path. This fixes the path issue in the `render-configs.sh` script.

### DIFF
--- a/render_configs.sh
+++ b/render_configs.sh
@@ -15,7 +15,7 @@ if ! [ -x "$(command -v vault)" ]; then
 fi
 
 $VAULT_COMMAND -field=data -format=json "secret/dsde/firecloud/$ENV/common/firecloud-account.json" >"$INTEGRATION_OUTPUT_LOCATION/user-delegated-sa.json"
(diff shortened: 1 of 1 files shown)


pshapiro4broad (approved):
This looks good. Is there a PR for helm as well? The swagger ID for our dev/perf/etc deployments also appears to be incorrect.

fboulnois:
> This looks good. Is there a PR for helm as well? The swagger ID for our dev/perf/etc deployments also appears to be incorrect.

I think the paths in the other environments are correct: https://github.com/broadinstitute/terra-helmfile/blob/master/charts/catalog/templates/secrets.yaml#L126

It's still possible that the swagger tokens are out-of-date / not properly enabled.

s-rubenstein (approved):
LGTM

PR #69, opened 2022-04-19
Add a gradle subproject for scripts.

pshapiro4broad:
Add a gradle subproject for scripts. This migrates the `shellcheck` and `black` operations from GitHub actions to gradle tasks, which makes it easier to run these operations locally.

With this change, `black` and `shellcheck` are triggered via the `check` task, which is run by `test`. So running `build` at the root level will also run `black` and `shellcheck` on the scripts subproject, if that project is out of date.

To use this, you must have `shellcheck` installed at `/usr/local/bin/shellcheck` (which is where `brew install shellcheck` puts it), and `black` must be installed and on your `PATH` (using `pip install black` should do this).

Another nice thing about having a subproject for scripts is it makes it easier to configure IntelliJ to support python editing features for the `.py` files.

This change also adds python settings to `.editorconfig`.


scripts/build.gradle
```diff
@@ -0,0 +1,63 @@
+import java.nio.file.Files
```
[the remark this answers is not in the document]
pshapiro4broad: They require that they're installed before running. I could configure shellcheck to run via docker, but that is slow and requires that you're running docker. So having shellcheck installed seems less burdensome.

To complete this PR, I'd need to add instructions on how to install these tools to the README, and update the GitHub workflows to install these tools so it can be run on PRs.

Alternatively, I could check it in without adding the `check dependsOn` below. That would be an incremental step and wouldn't require all users to install black and shellcheck and wouldn't require changing the GitHub workflows. I'd still update the README with instructions on how to set up the tools and run them locally.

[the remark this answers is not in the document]
fboulnois: We could add `shellcheck` and `black` to the `brewfile`. That would simplify stuff a lot.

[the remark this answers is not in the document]
pshapiro4broad: I added `shellcheck` but `black` requires `pip install` so I added a directive for that. Maybe we need a proper section on how to set up python for this project.

.github/workflows/build-and-test.yml
```diff
@@ -13,31 +13,29 @@ jobs:
...
     steps:
       - uses: actions/checkout@v2
+      - name: Link shellcheck into expected location
+        run: sudo ln -s $(which shellcheck) /usr/local/bin/shellcheck
+      - name: Set up Python
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.9
+      - name: Install Black
+        run: pip install black
```
fboulnois, line 23: I'm not sure we should block the Java build on installing and running `shellcheck`, `Python`, and `black` first. My preference is to keep the linting step and not make `shellcheck` / `black` a mandatory part of the Java build.

fboulnois (commented):
A few suggestions:

Testing.md
```diff
@@ -0,0 +1,65 @@
+# How to create a catalog entry in the terra data catalog
```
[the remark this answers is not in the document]
pshapiro4broad: Whoops, I committed this file by mistake. These are notes on testing I wrote up a while ago but hadn't finished. I can try to finish them and move them to a better place if it's helpful to others.

docs/getting-started.md
```diff
@@ -124,6 +124,9 @@ gcloud auth configure-docker
 # ensure that git-secrets patterns are installed
 git clone https://github.com/broadinstitute/dsp-appsec-gitsecrets-client.git
 ./dsp-appsec-gitsecrets-client/gitsecrets.sh
+
+# Install black, the Python code formatter
+pip install black
```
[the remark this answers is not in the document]
pshapiro4broad: Nice, I didn't realize the brew package works that way. I'll give it a try and update the docs/brewfile.

Testing.md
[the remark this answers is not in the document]
pshapiro4broad: I'm going to remove it for now.

.github/workflows/build-and-test.yml
pshapiro4broad: I don't see this as blocking the java build; my thinking is that the top level build is now building all gradle subprojects. This is similar to the `unit-tests-and-sonarqube` step, which runs the `test` task for all subprojects.

Running all builds in a single step means that we don't need to add a new step for each gradle subproject. It does have the downside that a lint error in the python code will block the java unit tests from running. The alternative is that the workflow code will need to be very specific about java build vs everything else, and java test vs everything else.

pshapiro4broad: Performance-wise this change has a negligible effect on the `build` job; the job completes in less than two minutes.

pshapiro4broad:
Example build with shellcheck failure: https://github.com/DataBiosphere/terra-data-catalog/runs/6212722425?check_suite_focus=true

.github/workflows/build-and-test.yml
pshapiro4broad: > I think it does effectively block the Java build: the action steps run sequentially and require the installation of shellcheck in a specific place and black before the build can proceed, and then the build uses those tools to run the linting. Ideally we would be able to do this in parallel to the main build, which is similar to what the linters job did beforehand. 

This is true, but what would be the gain of this change? Right now the the performance difference is negligible (less than 1 minute). An error in one area (like shellcheck) might block detection of errors in another area (like black) but once a user knows there's a compilation error, they can iterate locally. The downside is that the GitHub workflow would be more complex, which means harder to maintain. Keeping the workflow less complex seems more important to me than making these steps parallel. It's possible gradle supports this kind of parallelism, so there might be a way to fix this problem without adding complexity.

> requiring shellcheck to exist at a particular location for the build to run seems unusual to me.

This is a separate issue. The location is configurable but I don't know if there's a way to find it on the path and then pass that location to the plugin. The plugin can run `shellcheck` inside docker, which avoids this issue, but there's an overhead to launching docker that I'd like to avoid if possible. Looking at the [docs](https://github.com/felipefzdz/gradle-shellcheck-plugin#the-shellcheck-plugin), I think using `SHELLCHECK_PATH` would avoid the need to link shellcheck into place, so I can look into that too.

> I agree that running all builds in a single step is a convenience as a user, but I don't necessarily see that as an advantage in a GitHub Action where we may need to build specific parts of the project for specific cases (client vs server code, different types of scripts, etc)

That's true, it's not necessarily an advantage, although so far we don't have a need to build specific parts in our action.

> Additionally, as a user if I am making Java code change I shouldn't need to scan all my Python and shell scripts if those haven't changed

Gradle caches the results of a task run and shouldn't re-run the tasks unless their sources have changed. So this change shouldn't affect local development in this way.

.github/workflows/build-and-test.yml
```diff
@@ -13,31 +13,29 @@ jobs:
...
     steps:
       - uses: actions/checkout@v2
+      - name: Link shellcheck into expected location
+        run: sudo ln -s $(which shellcheck) /usr/local/bin/shellcheck
+      - name: Set up Python
+        uses: actions/setup-python@v1
+        with:
+          python-version: '3.10'
+      - name: Install Black
+        run: pip install black
```
fboulnois, line 23:
```suggestion
      - name: Set up Python
        uses: actions/setup-python@v1
        with:
          python-version: '3.10'
      - name: Install black and link shellcheck into expected location
        run: |
          pip install black
          sudo ln -s $(which shellcheck) /usr/local/bin/shellcheck
```

These steps can potentially be simplified as follows.

fboulnois (approved):
I'm still not totally sure about the action changes, but in the interest of not blocking merging and since the rest looks good to me I approve.

scripts/hca/ingest-hca-collection.py
```diff
@@ -1,11 +1,16 @@
...
-from urllib.request import Request, urlopen
+import json
+import os
+
+import requests
 from tqdm import tqdm
 
 upsert_url = "https://catalog.dsde-dev.broadinstitute.org/api/v1/datasets"
-policy_url = "https://jade.datarepo-dev.broadinstitute.org/api/repository/v1/snapshots/{id}/policies/steward/members"
+policy_url = (
```
s-rubenstein, line 10: Why does this one need to be paren'd and the one above isnt? Is it a line length thing?

s-rubenstein (approved):
Overall looks good to me. Should there be a ticket number for the PR? Also, two questions (but don't need to block the PR on them.)
1) I noticed a bunch of changes to python code here, I assume that's from Black? Otherwise I would probably say we should spin that out into separate PRs relating to the code cleanup, just to keep the diffs small and well contained.
2) Just out of curiosity, based on the conversations before, are there ways to set custom error messages with gradle? I was particularly wondering if, when we were using Pip to install Black, we could have it spit out a clean error message with the instructions for installing black if the error was that it wasn't found. Fully just for my own understanding though.

scripts/hca/ingest-hca-collection.py
pshapiro4broad: Sorry, I meant to revert the changes to this file. These changes are both some improvements and changes from updating to conform to PEP8 (which has shorter lines, among other things). Black is different than PEP8 so these changes aren't necessary for this PR.

pshapiro4broad:
> Should there be a ticket number for the PR?

I didn't think it was necessary since it's a change to how our checks are run (preserving current behavior, not adding new checks) and it was done "off the clock". But the reviewing has taken up more time than I thought it would, so maybe it should've had one.

> I noticed a bunch of changes to python code here, I assume that's from Black? Otherwise I would probably say we should spin that out into separate PRs relating to the code cleanup, just to keep the diffs small and well contained.

I agree, I've reverted the file with the most python changes. I was thinking we'd want to follow PEP8 too but that's something to discuss as a team before making any changes.

> Just out of curiosity, based on the conversations before, are there ways to set custom error messages with gradle? I was particularly wondering if, when we were using Pip to install Black, we could have it spit out a clean error message with the instructions for installing black if the error was that it wasn't found. Fully just for my own understanding though.

There's nothing built into gradle for this, as far as I can tell. It can be done by creating a custom task (which calls `System.exec`, or searches all locations in `$PATH`) and then making the downstream task depend on it. I think it would be hard to make it cache properly since the build environment isn't declared as a task input

PR #70, opened 2022-04-20
[DC-283] Working table preview endpoint

s-rubenstein:


service/src/test/java/bio/terra/catalog/service/DatasetServiceTest.java
```diff
@@ -190,11 +192,44 @@ void getDatasetPreviewTables() {
...
+                                List.of(
+                                    new ColumnModel()
+                                        .datatype(TableDataType.INTEGER)
+                                        .name("column a"))))));
+    when(datarepoService.getPreviewTable(user, tdrDataset.storageSourceId(), tableName))
+        .thenReturn(new SnapshotPreviewModel().result(List.of()));
+    DatasetPreviewTable datasetPreviewTable =
+        datasetService.getDatasetPreview(user, tdrDataset.id(), tableName);
+    assertThat(datasetPreviewTable.getRows().size(), is(0));
+    assertThat(datasetPreviewTable.getColumns().size(), is(1));
```
pshapiro4broad, line 226: There are specific asserts for collection size
```suggestion
    assertThat(datasetPreviewTable.getRows(), empty());
    assertThat(datasetPreviewTable.getColumns(), hasSize(1));
```

service/src/main/java/bio/terra/catalog/service/DatasetService.java
```diff
@@ -161,4 +195,18 @@ private List<TableMetadata> convertDatarepoTablesToCatalogTables(
                     .hasData(tableModel.getRowCount() > 0))
         .toList();
   }
+
+  private ColumnModel convertDatarepoColumnModelToCatalogColumnModel(
```
pshapiro4broad, line 199: Can this can be `static`? Declaring it as `static` would make it clearer that this doesn't rely on class state.

common/openapi.yml
```diff
@@ -297,6 +321,36 @@ components:
...
+        datatype:
+          $ref: '#/components/schemas/TableDataType'
+        array_of:
+          type: boolean
+          description: if true, make this column an array of type datatype.
+          default: false
+
+    TableDataType:
+      type: string
+      enum: [ string, boolean, bytes, date, datetime, dirref, fileref, float, float64, integer, int64, numeric, record, text, time, timestamp ]
```
pshapiro4broad, line 350: I think I mentioned this before, but I'm not sure about the use of including column types in the response. The UI doesn't currently use types, and we don't know what types other storage systems support. When we add new systems, we'll have to map that systems's types to our types. I think it would be better to only include data the UI needs to limit the the scope of our API, and to lower the cost of adding new storage systems.

common/openapi.yml
```diff
@@ -297,6 +321,36 @@ components:
...
+    ColumnModel:
+      type: object
+      properties:
+        name:
+          type: string
+        datatype:
+          $ref: '#/components/schemas/TableDataType'
+        array_of:
+          type: boolean
+          description: if true, make this column an array of type datatype.
```
pshapiro4broad, line 345: This sounds like the description for creation, not for an API result 
```suggestion
          description: if true, this column is an array of data
```

common/openapi.yml
```diff
@@ -297,6 +321,36 @@ components:
...
+          default: false
```
pshapiro4broad, line 346: I don't think a default value makes sense for a schema that's only used for a response. What would it be used for?

pshapiro4broad (approved):
Looks good, just a few last comments to address.

service/src/test/java/bio/terra/catalog/api/DatasetApiControllerTest.java
```diff
@@ -145,17 +149,43 @@ void createDataset() throws Exception {
   }
 
   @Test
-  void getDatasetPreviewTables() throws Exception {
+  void listDatasetPreviewTables() throws Exception {
     var datasetId = new DatasetId(UUID.randomUUID());
+    var tableName = "a";
```
fboulnois, line 154: maybe for consistency with the other test rename this to `table`?

service/src/test/java/bio/terra/catalog/service/DatasetServiceTest.java
```diff
@@ -190,11 +192,40 @@ void getDatasetPreviewTables() {
...
     assertThat(results.getTables().size(), is(1));
     assertThat(results.getTables().get(0), isA(TableMetadata.class));
     assertThat(results.getTables().get(0).isHasData(), is(true));
   }
+
+  @Test
+  void getDatasetPreviewTable() {
+    var tdrDataset =
+        new Dataset(dataset.id(), sourceId, StorageSystem.TERRA_DATA_REPO, metadata, null);
+    var tableName = "tableName";
```
fboulnois, line 206:
```suggestion
    var tableName = "table";
```

fboulnois (approved):
A minor suggestion to rename things for consistency, but not a blocker:

PR #71, opened 2022-04-21
DC-286: publish perf test results to QA dashboard

pshapiro4broad:
This migrates changes from QA-1765 branch to current code base. This adds a command to the nightly builds that uploads test results into  GS bucket.

### DIFF
--- a/.github/workflows/nightly-tests.yml
+++ b/.github/workflows/nightly-tests.yml
@@ -58,6 +58,18 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
+
+      - name: Alternative way of getting terra-helmfile repo version files
+        run: |
+          curl -H 'Authorization: token ${{ secrets.BROADBOT_TOKEN }}' \
+            -H 'Accept: application/vnd.github.v3.raw' \
+            -L https://api.github.com/repos/broadinstitute/terra-helmfile/contents/versions/app/dev.yaml \
+            --create-dirs -o "integration/terra-helmfile/versions/app/dev.yaml"
+          curl -H 'Authorization: token ${{ secrets.BROADBOT_TOKEN }}' \
+            -H 'Accept: application/vnd.github.v3.raw' \
+            -L https://api.github.com/repos/broadinstitute/terra-helmfile/contents/environments/live/perf.yaml \
+            --create-dirs -o "integration/terra-helmfile/environments/live/perf.yaml"
+
       - name: Set up JDK
         uses: actions/setup-java@v2
         with:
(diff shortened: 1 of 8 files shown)


render_configs.sh
```diff
@@ -2,7 +2,7 @@ ENV=${1:-dev}
 VAULT_TOKEN=${2:-$(cat "$HOME"/.vault-token)}
 
 VAULT_ADDR="https://clotho.broadinstitute.org:8200"
-COMMON_VAULT_PATH="secret/dsde/terra/kernel/$ENV/common"
+COMMON_VAULT_PATH="secret/dsde/terra/kernel/perf/common"
```
s-rubenstein, line 5: Why this change? I would expect us to still want render_configs to be generic.

pshapiro4broad: As I understand it, the only testrunner account that exists is the perf one. Before we were only using the perf account, but the names were written as if it was a generic SA that existed in all environments. This change makes the code more explicit that only `perf` is supported.

I'll update the environment variable name to also be perf specific.

.github/workflows/nightly-tests.yml
```diff
@@ -58,6 +58,18 @@ jobs:
...
+      - name: Get the helm chart versions for the perf env
+        run: |
+          curl -H 'Authorization: token ${{ secrets.BROADBOT_TOKEN }}' \
+            -H 'Accept: application/vnd.github.v3.raw' \
+            -L https://api.github.com/repos/broadinstitute/terra-helmfile/contents/versions/app/dev.yaml \
+            --create-dirs -o "integration/src/main/resources/rendered/dev.yaml"
+          curl -H 'Authorization: token ${{ secrets.BROADBOT_TOKEN }}' \
+            -H 'Accept: application/vnd.github.v3.raw' \
+            -L https://api.github.com/repos/broadinstitute/terra-helmfile/contents/environments/live/perf.yaml \
+            --create-dirs -o "integration/src/main/resources/rendered/perf.yaml"
```
fboulnois, line 71: There might be a difference between what is in the helm charts and what is deployed to the dev / perf env. Should we curl the dev / perf version endpoints instead?

pshapiro4broad:
If you are on the VPN, you can see the result of running the nightly perf workflow here: https://trdash.dsp-eng-tools.broadinstitute.org/. Select the date `04/26/22` and then select `catalog-perf` from the `Env type` menu.

fboulnois (commented):
The main change I see is uploading to the bucket. A suggestion:

s-rubenstein (approved):
LGTM, pending @fboulnois 's question

.github/workflows/nightly-tests.yml
pshapiro4broad: I totally agree! However the current QA dashboard requires a helm chart version which our version endpoint doesn't return. I've talked with the QA team and the plan is that the next version of testrunner will support using the `version` endpoint of a service to get this information.

pshapiro4broad: It would be much better to get the version information from this endpoint, and would allow us to remove this `curl` code. But for now this is the "best" way to do this.

fboulnois (approved):

ichengchang (approved):
Looks good!

PR #72, opened 2022-04-22
DC-319: Update perf catalog tests to run using dev TDR

pshapiro4broad:
Update the perf catalog tests to use dev TDR instead of perf TDR. The perf TDR environment isn't stable and can't currently be used for catalog tests.


fboulnois (approved):

s-rubenstein (approved):
LGTM

PR #74, opened 2022-04-29
DC 297 add permissions tests

tcjiang9:


integration/src/main/java/scripts/testscripts/DatasetPermissionOperations.java
```diff
@@ -0,0 +1,238 @@
...
+            .getDefaultProfileId();
+    adminClient = new CatalogClient(server, adminUser);
+    adminDatasetsApi = new DatasetsApi(adminClient);
+  }
+
+  @Override
+  public void userJourney(TestUserSpecification testUser) throws Exception {
+    DatasetsApi userDatasetsApi = new DatasetsApi(new CatalogClient(server, regularUser));
+
+    testReaderPermissions(userDatasetsApi);
```
pshapiro4broad, line 74: Can the same snapshot can be used for all of these tests, since they aren't running in parallel? Each test would have to ensure that the permissions are in the expected state before running, though. Since snapshots are so expensive to make, it would be worth recycling one if possible.

pshapiro4broad (commented):
First round of comments / feedback.

integration/src/main/java/scripts/testscripts/DatasetPermissionOperations.java
```diff
@@ -0,0 +1,238 @@
...
+  @Override
+  public void cleanup(List<TestUserSpecification> testUsers) throws Exception {
+    if (!snapshotIds.isEmpty()) {
+      for (var snapshotId : snapshotIds) {
+        try {
+          adminSnapshotsApi.deleteSnapshot(snapshotId);
+        } catch (Exception e) {
+          var datarepoClient = new DatarepoClient(server, regularUser);
+          var snapshotsApi = new SnapshotsApi(datarepoClient);
+          snapshotsApi.deleteSnapshot(snapshotId);
```
pshapiro4broad, line 225: This case can be avoided if you first set permission on the snapshot to allow the admin to delete it.

integration/src/main/java/scripts/testscripts/DatasetPermissionOperations.java
```diff
@@ -0,0 +1,238 @@
...
+  @Override
+  public void setup(List<TestUserSpecification> testUsers) throws Exception {
+    for (TestUserSpecification testUser : testUsers) {
+      if ("datacatalogadmin@test.firecloud.org".equals(testUser.userEmail)) {
+        adminUser = testUser;
+      }
+      if ("datacataloguser@test.firecloud.org".equals(testUser.userEmail)) {
+        regularUser = testUser;
+      }
+    }
```
pshapiro4broad, line 57: As it turns out, the client APIs won't throw an error if the user argument is `null`; they instead create a no-auth client instead of an auth client. So after the `for` loop you need a couple asserts like:
```
    assertThat(adminUser, notNullValue());
    assertThat(regularUser, notNullValue());
```

integration/src/main/java/scripts/testscripts/DatasetPermissionOperations.java
```diff
@@ -0,0 +1,238 @@
...
+          var snapshotsApi = new SnapshotsApi(datarepoClient);
+          snapshotsApi.deleteSnapshot(snapshotId);
+        } finally {
+          log.info("deleted snapshot " + snapshotId);
+        }
+      }
+    }
+    if (!datasetIds.isEmpty()) {
+      for (var datasetId : datasetIds) {
+        adminDatasetsApi.deleteDataset(datasetId);
```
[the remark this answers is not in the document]
tcjiang9: Wait I got this from SnapshotDatasetOperations:
```
  @Override
  public void cleanup(List<TestUserSpecification> testUsers) throws Exception {
    if (snapshotId != null) {
      snapshotsApi.deleteSnapshot(snapshotId);
      log.info("deleted snapshot " + snapshotId);
    }
    if (datasetId != null) {
      datasetsApi.deleteDataset(datasetId);
      log.info("deleted dataset " + datasetId);
    }
  }

```
which you wrote. 
Why is it different here?

integration/src/main/java/scripts/testscripts/DatasetPermissionOperations.java
```diff
@@ -0,0 +1,238 @@
...
+    var request =
+        new CreateDatasetRequest()
+            .catalogEntry(METADATA)
+            .storageSourceId(snapshotId.toString())
+            .storageSystem(StorageSystem.TDR);
+    // Verify admin can create a dataset on the user snapshot
+    var datasetId = adminCreateDataset(request);
+    // But admin cannot access the underlying user snapshot
+    try {
+      adminSnapshotsApi.retrieveSnapshot(snapshotId, Collections.emptyList());
```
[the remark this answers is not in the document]
tcjiang9: Yeah this doesn't make a lot of sense to me either but it was in the ticket description
"An admin can create/update/delete a catalog entry for a user’s snapshot but cannot access the underlying data"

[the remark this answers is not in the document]
pshapiro4broad: > but cannot access the underlying data

The way to test this using catalog APIs is to use the preview endpoints, either `listDatasetPreviewTables()` or `getDatasetPreviewTable()` should trigger an error.

integration/src/main/java/scripts/testscripts/DatasetPermissionOperations.java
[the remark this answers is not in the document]
pshapiro4broad: Yeah it's wrong, I thought I had fixed the order but I guess that change didn't get committed. It turns out either way will work because the auth check is done without calling TDR since the user is an admin user. But conceptually the dataset should be removed from catalog before the snapshot is deleted from TDR. If you could fix the other case too, that would be good.

integration/src/main/java/scripts/testscripts/DatasetPermissionOperations.java
```diff
@@ -0,0 +1,238 @@
...
+      if ("datacatalogadmin@test.firecloud.org".equals(testUser.userEmail)) {
+        adminUser = testUser;
+      }
+      if ("datacataloguser@test.firecloud.org".equals(testUser.userEmail)) {
+        regularUser = testUser;
+      }
+    }
+    var datarepoClient = new DatarepoClient(server, adminUser);
+    adminSnapshotsApi = new SnapshotsApi(datarepoClient);
+    defaultProfileId =
```
fboulnois, line 60: It might be worth asserting this is non-null, too.

integration/src/main/java/scripts/testscripts/DatasetPermissionOperations.java
```diff
@@ -0,0 +1,238 @@
...
+            .addContentsItem(
+                new SnapshotRequestContentsModel()
+                    .datasetName(TEST_DATASET_NAME)
+                    .mode(SnapshotRequestContentsModel.ModeEnum.BYFULLVIEW));
+    var createJobId = snapshotsApi.createSnapshot(request).getId();
+
+    var jobApi = new JobsApi(datarepoClient);
+    JobModel job;
+    do {
+      job = jobApi.retrieveJob(createJobId);
```
fboulnois, line 202: Should we sleep here to not burn CPU and network requests?

integration/src/main/resources/testusers/admin.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "Andy Anteater",
+  "name": "Admin User",
```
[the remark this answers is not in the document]
fboulnois: I agree with @tcjiang9 that the roles should be obvious and weren't at first glance. Perhaps Andy 'Admin' Anteater?

integration/src/main/resources/testusers/user.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "Ursula Upton",
+  "name": "Regular User",
```
fboulnois, line 2:
```suggestion
  "name": "Ursula 'User' Upton",
```

fboulnois (commented):
I largely agree with Phil's suggestions. Here are a few additional items:

integration/src/main/resources/configs/integration/DatasetPermissionOperations.json
```diff
@@ -0,0 +1,17 @@
...
+  "serverSpecificationFile": "catalog-local.json",
+  "kubernetes": {},
+  "application": {},
+  "testScripts": [
+    {
+      "name": "DatasetPermissionOperations",
+      "numberOfUserJourneyThreadsToRun": 1,
+      "userJourneyThreadPoolSize": 2,
+      "expectedTimeForEach": 5,
+      "expectedTimeForEachUnit": "MINUTES"
```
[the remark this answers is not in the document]
tcjiang9: do you mean move snapshot creation to setup, and just have one snapshot shared by all the test helper functions?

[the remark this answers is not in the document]
pshapiro4broad: Yes, I was thinking you could narrow it down to two snapshots, one created by `admin` and one created by `user`.

You could probably use a single one since `admin` has `SET_POLICY` it can add/remove all roles before running a test. But maybe we want to let TDR manage the owner setting and not modify that in our test?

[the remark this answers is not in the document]
pshapiro4broad: But also yes, move the snapshot creation into `setup()`.

integration/src/main/java/scripts/testscripts/DatasetPermissionOperations.java
pshapiro4broad: Is it possible for a dataset to not have a default profile ID?

integration/src/main/java/scripts/testscripts/DatasetPermissionOperations.java
pshapiro4broad: That's a good point, this is code copied from `SnapshotDatasetOperations.setup()` but the same is true there; the code should be sleeping as we know it takes ~60s for this to complete. Maybe a 10 second sleep would be good?

integration/src/main/resources/testusers/admin.json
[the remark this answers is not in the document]
pshapiro4broad: It's fine to rename them if it makes the intent clearer. Actually maybe what we should do here is make it
```suggestion
  "name": "admin",
```

and then in the code check for `user.name.equals("admin")` instead of checking the email address. That would be better I think.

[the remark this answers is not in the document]
tcjiang9: I'd prefer to check for the email here, since that's what matters when it comes to the roles

integration/src/main/java/scripts/testscripts/DatasetPermissionOperations.java
tcjiang9: I'd prefer to just keep this here to reduce the amount of state that tests have to manage

pshapiro4broad (approved):
Looks good, I would add one additional check to the discoverer test.

Also I see what you mean about putting all the tests in a single script means that you can't run an individual test anymore. I was thinking as we add more tests I wanted to avoid having a new test config for each test. But maybe that's OK and the ability to run a specific test is more important than trying to limit the number of json files in our repo.

fboulnois (approved):
lgtm

integration/src/main/java/scripts/common/ApiHelpers.java
```diff
@@ -0,0 +1,43 @@
...
+import bio.terra.datarepo.api.SnapshotsApi;
+import bio.terra.datarepo.client.ApiClient;
+import bio.terra.datarepo.client.ApiException;
+import bio.terra.datarepo.model.JobModel;
+import bio.terra.datarepo.model.SnapshotRequestModel;
+import java.util.Map;
+import java.util.UUID;
+import java.util.concurrent.TimeUnit;
+
+public class ApiHelpers {
```
pshapiro4broad, line 13: Since each `ApiClient` is specific to a service, I would put this code inside `DatarepoClient` (vs a more generally named `ApiHelpers`)

integration/src/main/java/scripts/common/ApiHelpers.java
```diff
@@ -0,0 +1,43 @@
...
+  public static Map<Object, Object> waitForJobSuccess(ApiClient datarepoClient, String jobId)
```
pshapiro4broad, line 14: I think this can become a non-static method on `DatarepoClient`

integration/src/main/java/scripts/common/ApiHelpers.java
```diff
@@ -0,0 +1,43 @@
...
+    } while (job.getJobStatus() == JobModel.JobStatusEnum.RUNNING);
+
+    if (job.getJobStatus() != JobModel.JobStatusEnum.SUCCEEDED) {
+      throw new RuntimeException("Job failed: " + job);
+    }
+    return (Map<Object, Object>) jobApi.retrieveJobResult(jobId);
+  }
+
+  public static UUID synchronousCreateSnapshot(
+      SnapshotsApi snapshotsApi, SnapshotRequestModel request)
```
pshapiro4broad, line 30: With this code refactored, you could simplify the caller even more by only passing in the field of `request` that are different for each call. I think that comes down to only `profileID` (or make it accept `Dataset`).

Also IMO I would remove a prefix like `synchronous` from these names unless we have an async version and need to disambiguate.

integration/src/main/java/scripts/common/ApiHelpers.java
```diff
@@ -0,0 +1,43 @@
...
+    JobModel job;
+    do {
+      job = jobApi.retrieveJob(jobId);
+      TimeUnit.SECONDS.sleep(5);
+    } while (job.getJobStatus() == JobModel.JobStatusEnum.RUNNING);
+
+    if (job.getJobStatus() != JobModel.JobStatusEnum.SUCCEEDED) {
+      throw new RuntimeException("Job failed: " + job);
+    }
+    return (Map<Object, Object>) jobApi.retrieveJobResult(jobId);
```
pshapiro4broad, line 26: I believe this will generate a warning about an unchecked cast. We want the code to compile without any warning, so a `SupressWarnings` is needed.
```suggestion
    @SuppressWarnings("unchecked")
    return (Map<Object, Object>) jobApi.retrieveJobResult(jobId);
```

integration/src/main/java/scripts/testscripts/DatasetPermissionOperations.java
```diff
@@ -0,0 +1,210 @@
...
+    }
+    for (var snapshotId : snapshotIds) {
+      try {
+        ApiHelpers.synchronousDeleteSnapshot(adminSnapshotsApi, snapshotId);
+      } catch (Exception e) {
+        ApiHelpers.synchronousDeleteSnapshot(userSnapshotsApi, snapshotId);
+      } finally {
+        log.info("deleted snapshot " + snapshotId);
+      }
+    }
```
pshapiro4broad, line 208: Now that there are only two snapshots, `adminTestSnapshotId` and `userTestSnapshotId`, we no longer need `snapshotIds` and this loop.
```suggestion
    ApiHelpers.synchronousDeleteSnapshot(adminSnapshotsApi, adminTestSnapshotId);
    log.info("deleted snapshot " + adminTestSnapshotId);
    ApiHelpers.synchronousDeleteSnapshot(userSnapshotsApi, userTestSnapshotId);
    log.info("deleted snapshot " + userTestSnapshotId);
```

We could remove the `log`s too; they were nice when the tests were first being debugged but are probably less useful now. Since we're always using the same dataset for all tests, it wouldn't be hard to find the snapshots from that dataset if we needed to.

integration/src/main/java/scripts/testscripts/SnapshotDatasetOperations.java
```diff
@@ -53,7 +54,13 @@ public class SnapshotDatasetOperations extends TestScript {
...
   @Override
   public void setup(List<TestUserSpecification> testUsers) throws Exception {
-    DatarepoClient datarepoClient = new DatarepoClient(server, testUsers.get(0));
+    DatarepoClient datarepoClient = null;
+    for (TestUserSpecification testUser : testUsers) {
+      if (testUser.userEmail.equals(ADMIN_EMAIL)) {
+        datarepoClient = new DatarepoClient(server, testUser);
+      }
+    }
+    assertNotNull(datarepoClient);
```
pshapiro4broad, line 63: This change is unnecessary as this test only has a single element in `testUsers` and it's always the admin user. If we need to add more users in the future, we can update the code then to handle this case.
```suggestion
    DatarepoClient datarepoClient = new DatarepoClient(server, testUsers.get(0));
 ```

**Review by pshapiro4broad — APPROVED** [review]
Looks ok although I have a few more comments about the refactored code.

File: integration/src/main/java/scripts/common/ApiHelpers.java (hunk 13 above)
Comment (tcjiang9) [reply integration/src/main/java/scripts/common/ApiHelpers.java:30]: the async version would just be regular createSnapshot

File: integration/src/main/java/scripts/common/ApiHelpers.java (hunk 11 above)
Comment (tcjiang9) [reply integration/src/main/java/scripts/common/ApiHelpers.java:13]: if I move createSnapshot and deleteSnapshot to DatarepoClient I think it makes the code awkwardly organized, since DatarepoClient will have the create and delete functions while SnapshotsApi will still have addSnapshotPolicyMember and other API functions

File: integration/src/main/java/scripts/common/ApiHelpers.java (hunk 13 above)
Comment (pshapiro4broad) [reply integration/src/main/java/scripts/common/ApiHelpers.java:30]: Maybe, but we don't have an async version, and don't have a need for one in the future. So all `createSnapshot()` method calls in our code are sync. So there's no need to disambiguate sync from async in the method name.

File: integration/src/main/java/scripts/common/ApiHelpers.java (hunk 11 above)
Comment (pshapiro4broad) [reply integration/src/main/java/scripts/common/ApiHelpers.java:13]: That's true, they don't really belong in `DatarepoClient`. Could you subclass `SnapshotsApi` and add the methods to that? Something like

```
public class SnapshotsApi extends bio.terra.datarepo.api.SnapshotsApi {
  public SnapshotsApi(DatarepoClient apiClient) {
    super(apiClient);
  }

  public DatarepoClient getApiClient() {
    return (DatarepoClient) super.getApiClient();
  }

  public void synchronousDeleteSnapshot(UUID snapshotId) throws Exception {
    String deleteJobId = deleteSnapshot(snapshotId).getId();
    getApiClient().waitForJobSuccess(deleteJobId);
  }

  public UUID synchronousCreateSnapshot(SnapshotRequestModel request) throws Exception {
    String createJobId = createSnapshot(request).getId();
    Map<Object, Object> result = getApiClient().waitForJobSuccess(createJobId);
    return UUID.fromString((String) result.get("id"));
  }
}
```

And in `DatarepoClient`:
```
  public SnapshotsApi createSnapshotsApi() {
    return new SnapshotsApi(this);
  }
```

pshapiro4broad: (inheritance makes this easier to write but encapsulation may be a better approach)

PR #75, opened 2022-05-02
[DC-210] Added mapping initial readme for converting workspaces to catalog snapshots

lu-c:
Want to make a draft PR to see if this looks okay to the team!


scripts/workspace-conversion/README.md
```diff
@@ -0,0 +1,60 @@
...
+# Value mappings
+To run the conversion from workspace to dataset, here is a current mapping scheme:
+
+| Attribute | Rawls Workspace Path | Data Catalog Snapshot Path |
+| --------- | -------------------- | -------------------------- |
+| Cohort Phenotype/Indication (Disease Ontology)  | `library:diseaseOntologyLabel` |  |
+| No. of Subjects | `library:numSubjects` | `counts.donors` |
+| Data Category | `library:dataCategory`<br>ex: "Simple Nucleotide Variation, Copy Number Variation, Expression Quantification, DNA-Methylation, Clinical phenotypes, Biosample metadata" |  |
+| Experimental Strategy | `library:datatype`<br>ex: "Whole Exome, Genotyping Array, RNA-Seq, miRNA-Seq, Methylation Array, Protein Expression Array" | `prov:wasGeneratedBy.3.TerraCore:hasAssayCategory`<br>Note: RNA-Seq would be found in this field in existing data catalog entries, but I don't know if all of these values would also fit in the same one. |
+| Data Use Limitation | `library:dataUseRestriction` | `TerraDCAT_ap:hasDataUsePermission.0`<br>Not a perfect mapping, we will need to normalize "General Research Use" to "TerraCore:NoRestriction" |
```
fboulnois, line 36: Yes, probably an enum of values would be useful here.

scripts/workspace-conversion/README.md
```diff
@@ -0,0 +1,60 @@
...
+| Cohort Phenotype/Indication | `library:indication`<br>ex: "Mesothelioma" | `samples.disease.0` |
+| Cohort Name | `library:datasetName` | `dct:title` |
+| Dataset Version | `library:datasetVersion` |  |
```
fboulnois, line 39: I'm surprised we don't have a dataset version, but you're right! I think we should include one since we display it in the UI.

fboulnois, line 39: I'm surprised we don't have a dataset version, but you're right! I think we should include one since we display it in the UI.

scripts/workspace-conversion/README.md
```diff
@@ -0,0 +1,60 @@
...
+| Cohort Description | `library:datasetDescription` | `dct:description` |
+| Dataset Owner | `library:datasetOwner`<br>ex: "NCI" | `TerraDCAT_ap:hasDataCollection.0.dct:identifier`<br>If we have enough information, we can also build:<br> `TerraDCAT_AP:hasDataCollection.0.dct:publisher: National Cancer Institute`<br>`TerraDCAT_AP:hasDataCollection.0.dct:title: National Cancer Institute` |
+| Dataset Custodian | `library:datasetCustodian`<br>ex: "dbGAP" | `TerraDCAT_ap:hasDataCollection.0.dct:identifier`<br>Note: This conflicts with "Dataset Owner", but I dont think we have another option for where to put this. |
```
fboulnois, line 42: I think the TIM seperates the owner from the publisher, so we might be able to use that.

fboulnois, line 42: I think the TIM seperates the owner from the publisher, so we might be able to use that.

scripts/workspace-conversion/README.md
```diff
@@ -0,0 +1,60 @@
...
+| Dataset Depositor | `library:datasetDepositor` | `contributors.0.contactName`<br>`contributors.0.correspondingContributor = true` |
+| Contact Email | `library:contactEmail` | `contributors.0.email` |
+| Research Institute | `library:institute` | `contributors.0.institution` |
+| Primary Disease Site | `library:primaryDiseaseSite`<br>ex: "Pleura" | `samples.disease.0`?<br>ex: "Brain Cancer" |
+| Project Name | `library:projectName`<br>ex: "TCGA" |  |
```
fboulnois, line 47: This will probably be the consortium in `hasDataCollection`.

fboulnois, line 47: This will probably be the consortium in `hasDataCollection`.

scripts/workspace-conversion/README.md
```diff
@@ -0,0 +1,60 @@
...
+| Genome Reference Version | `library:reference`<br>ex: "GRCh37/hg19" |  |
+| Data File Formats | `library:dataFileFormats`<br>ex: "TXT, MAF" | `files.0.dcat:mediaType`<br>`files.0.count = 0`<br>`files.0.byteSize = 0`<br>Note: No way of knowing how many files match each file format |
```
fboulnois, line 49: Overall I agree, there's probably a subset we can convert though (but not necessarily for the first pass)

fboulnois, line 49: Overall I agree, there's probably a subset we can convert though (perhaps not necessarily for the first pass)

scripts/workspace-conversion/README.md
```diff
@@ -0,0 +1,60 @@
...
+| Profiling Protocol | `library:profilingProtocol` |  |
+| Depth of Sequencing Coverage (Average) | `library:coverage` |  |
+| Study Design | `library:studyDesign`<br>ex: "Tumor/Normal" | `samples.disease.0`<br>Note: concat value with "Primary Disease Site"? |
+| Cell Type | `library:cellType`<br>ex: "Primary tumor cell, Whole blood" | `samples.disease.0`<br>Note: concat value with "Primary Disease Site"? |
+| Reported Ethnicity | `library:ethnicity` |  |
+| Cohort Country of Origin | `library:cohortCountry`<br>ex: "USA" |  |
+| Requires External Approval | `library:requiresExternalApproval` |  |
+| library Metadata Schema Version Number | `library:lmsvn` |  |
+| Structured Data Use Limitations Version Number | `library:dulvn` |  |
+| ORSP Conset Code | `library:orsp` |  |
```
fboulnois, line 60: We have these for HCA data, but overall it's fair to say that we don't capture these in the TIM structure.

fboulnois (commented):
Looks good! I also recommend reaching out to Kathy.

scripts/workspace-conversion/README.md
lu-c: Since this isn't being used or set right now, I'm going to just set it to `dct:version` - I think dct means "data catalog"?

scripts/workspace-conversion/README.md
lu-c: Are these the tim fields?

```
{
   TerraDCAT_ap:hasDataCollection: {
      dct:creator: {foaf:name: "Human Cell Atlas"} <--- OWNER
      dct:identifier: "HCA" <---------- publisher?
      dct:publisher: "Human Cell Atlas" <------------ PUBLISHER
      dct:title: "Human Cell Atlas"
   },
   dct:creator: {foaf:name: "Human Cell Atlas"} <--- OWNER? (always seems to duplicate the owner value inside `TerraDCAT_ap:hasDataColletion.dct:creator`
}
```

scripts/workspace-conversion/README.md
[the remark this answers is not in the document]
lu-c: Yeah, we should see if we want to store these too, and whether or not it should belong in a specific separate field

scripts/workspace-conversion/README.md
```diff
@@ -0,0 +1,60 @@
+## Converting Workspace Dataset Attributes to Snapshot Data
```
pshapiro4broad, line 1: The target here isn't snapshot data but is catalog entry data. The distinction is that the current schema that the catalog UI accepts is not from TDR or snapshots but is a custom schema created for the data catalog. So this is an effort to map workspace properties to the catalog's schema.

A separate / associated effort would be to formally document what the data catalog's schema actually is. The schema is based on TIM (specifically [TerraDCAT_ap:Dataset](https://github.com/DataBiosphere/terra-interoperability-model/blob/master/releases/1.x/terra-core/TerraDCAT-AP.ttl#L167)

scripts/workspace-conversion/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+# Description
+This script takes the workspace dataset attributes and attempts to map them to similar fields in the data catalog service as snapshots.
```
pshapiro4broad, line 4:
```suggestion
This script takes the workspace dataset attributes and attempts to map them to similar fields in the data catalog service as catalog entries.
```

scripts/workspace-conversion/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+# Dataset Attributes
+An example workspace with attributes can be found [here](src/data/workspace-cleanAttributes.js.displayLibraryAttributes).
```
pshapiro4broad, line 7: Do you mean to link to https://github.com/DataBiosphere/terra-ui/blob/dev/src/data/workspace-attributes.js ?

scripts/workspace-conversion/README.md
```diff
@@ -0,0 +1,60 @@
...
+}
+```
+
+What is shown to the user in that table is the contents in the attributes that terra-ui has determined is important to display. These values are hardcoded at 
+`{terra-ui}/src/data/workspace-cleanAttributes.js.displayLibraryAttributes`
+
+# Value mappings
+To run the conversion from workspace to dataset, here is a current mapping scheme:
+
+| Attribute | Rawls Workspace Path | Data Catalog Snapshot Path |
```
pshapiro4broad, line 30:
```suggestion
| Attribute | Rawls Workspace Path | Data Catalog Entry Path |
```

scripts/workspace-conversion/README.md
lu-c: Gotcha, updating this title :)

scripts/workspace-conversion/README.md
lu-c: Oh thank you, this is definitely a better choice

scripts/workspace-conversion/README.md
```diff
@@ -0,0 +1,61 @@
...
+`{terra-ui}/src/data/workspace-cleanAttributes.js.displayLibraryAttributes`
+
+# Value mappings
+To run the conversion from workspace to dataset, here is a current mapping scheme:
+
+
+| i | Attribute | Rawls Workspace Path | Data Catalog Entry Path |
+| - | --------- | -------------------- | -------------------------- |
+| 1 | Cohort Phenotype/Indication (Disease Ontology)  | `library:diseaseOntologyLabel` | `samples.disease.0`<br>kathy: lists predefined terms, but forces them to stick with a specific vocabulary. The other phenotype field is the diseases |
+| 2 | No. of Subjects | `library:numSubjects` | `counts.donors`<br>Kathy: suggests leaving out this field because it is volatile and can easily become inaccurate |
```
kreinold, line 34: I would say that this should be calculated when the snapshot is made or calculated at run-time.

scripts/workspace-conversion/README.md
```diff
@@ -0,0 +1,61 @@
...
+| 20 | Profiling Protocol | `library:profilingProtocol` |  |
+| 21 | Depth of Sequencing Coverage (Average) | `library:coverage` |  |
+| 22 | Study Design | `library:studyDesign`<br>ex: "Tumor/Normal" | `samples.disease.0`<br>Note: concat value with "Primary Disease Site"? |
+| 23 | Cell Type | `library:cellType`<br>ex: "Primary tumor cell, Whole blood" | `samples.disease.0`<br>Note: concat value with "Primary Disease Site"? |
+| 24 | Reported Ethnicity | `library:ethnicity` |  |
+| 25 | Cohort Country of Origin | `library:cohortCountry`<br>ex: "USA" |  |
+| 26 | Requires External Approval | `library:requiresExternalApproval` |  |
+| 27 | library Metadata Schema Version Number | `library:lmsvn` |  |
+| 28 | Structured Data Use Limitations Version Number | `library:dulvn` |  |
+| 29 | ORSP Conset Code | `library:orsp` |  |
```
kreinold, line 61: TerraDCAT_ap:hasConsentGroup xsd:string

scripts/workspace-conversion/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+# Value mappings
+To run the conversion from workspace to dataset, here is a current mapping scheme:
+
+
+| i | Attribute | Rawls Workspace Path | Data Catalog Entry Path |
+| - | --------- | -------------------- | -------------------------- |
+| 1 | Cohort Phenotype/Indication (Disease Ontology)  | `library:diseaseOntologyLabel` | `samples.disease.0`<br>kathy: lists predefined terms, but forces them to stick with a specific vocabulary. The other phenotype field is the diseases |
+| 2 | No. of Subjects | `library:numSubjects` | `counts.donors`<br>Kathy: suggests leaving out this field because it is volatile and can easily become inaccurate |
+| 3 | Data Category | `library:dataCategory`<br>ex: "Simple Nucleotide Variation, Copy Number Variation, Expression Quantification, DNA-Methylation, Clinical phenotypes, Biosample metadata" | `dct:dataCategory`*<br>This field does not currently exist |
```
kreinold, line 35: Reviewing these examples it's pretty much a duplication of what's in Experimental Strategy (that we'll map to data modality).  I'm not sure if we need it at all but if there's concern about losing the information, please feel free to add this as a column. It won't be included in the model.

scripts/workspace-conversion/README.md
```diff
@@ -0,0 +1,61 @@
...
+| 4 | Experimental Strategy | `library:datatype`<br>ex: "Whole Exome, Genotyping Array, RNA-Seq, miRNA-Seq, Methylation Array, Protein Expression Array" | Kathy: Map this to data modality, talk to kathy about getting the mapping of terms |
+| 5 | Data Use Limitation | `library:dataUseRestriction` | `TerraDCAT_ap:hasDataUsePermission.0`<br>Not a perfect mapping, we will need to normalize "General Research Use" to "TerraCore:NoRestriction", preferably using the mapping found [here](https://github.com/DataBiosphere/terra-ui/blob/dev/src/pages/library/dataBrowser-utils.js#23) under `datasetReleasePolicies` |
```
kreinold, line 37: General Research Use is NOT the same as no restriction.  If the dataset has not adopted Data Use Ontology terms, I'm not sure we should use this field.  Can we discuss?

scripts/workspace-conversion/README.md
```diff
@@ -0,0 +1,61 @@
...
+| 6 | Cohort Phenotype/Indication | `library:indication`<br>ex: "Mesothelioma" | kathy: might not be useful, could consider pre-pending with source (ie: "TCGA" or "Anvil_phenotype") |
```
kreinold, line 38: Looking at the examples, I think this corresponds to disease.  We don't usually store that on a dataset because many datasets will have many diseases.  Again, we could add it as a field but long term, would we calculate it or do we need to provide a caveat that this data was submitted with the original dataset and may not include all diseases represented in the dataset.  You can add it as a field but it won't be part of TIM.

scripts/workspace-conversion/README.md
```diff
@@ -0,0 +1,61 @@
...
+| 7 | Cohort Name | `library:datasetName` | `dct:title` |
+| 8 | Dataset Version | `library:datasetVersion` | `dct:version` |
+| 9 | Cohort Description | `library:datasetDescription` | `dct:description` |
+| 10 | Dataset Owner | `library:datasetOwner`<br>ex: "NCI" | `TerraDCAT_ap:hasDataCollection.0.dct:identifier`<br>If we have enough information, we can also build:<br> `TerraDCAT_AP:hasDataCollection.0.dct:publisher: National Cancer Institute`<br>`TerraDCAT_AP:hasDataCollection.0.dct:title: National Cancer Institute` |
+| 11 | Dataset Custodian | `library:datasetCustodian`<br>ex: "dbGAP" | `TerraDCAT_ap:hasDataCollection.0.dct:identifier`<br>Note: This conflicts with "Dataset Owner", but I dont think we have another option for where to put this. |
```
kreinold, line 43: Not sure why this conflicts w/ Custodian.  Definition of owner is The entity responsible for making the item available and responsible for identifying authorized users of the Dataset or DataCollection.  A custodian is The entity operationally responsible for managing the item and responsible for authorizing users of the Dataset or DataCollection.

scripts/workspace-conversion/README.md
```diff
@@ -0,0 +1,61 @@
...
+| 12 | Dataset Depositor | `library:datasetDepositor` | `contributors.0.contactName`<br>`contributors.0.correspondingContributor = true` |
+| 13 | Contact Email | `library:contactEmail` | `contributors.0.email` |
+| 14 | Research Institute | `library:institute` | `contributors.0.institution` |
```
kreinold, line 46: prov:wasAssociatedWith xsd:string

scripts/workspace-conversion/README.md
```diff
@@ -0,0 +1,61 @@
...
+| 15 | Primary Disease Site | `library:primaryDiseaseSite`<br>ex: "Pleura" | `samples.disease.0`?<br>ex: "Brain Cancer" |
+| 16 | Project Name | `library:projectName`<br>ex: "TCGA" |  |
```
kreinold, line 48: prov:wasGeneratedBy xsd:string (name or URL)

scripts/workspace-conversion/README.md
```diff
@@ -0,0 +1,61 @@
...
+| 17 | Genome Reference Version | `library:reference`<br>ex: "GRCh37/hg19" |  |
```
kreinold, line 49: Genome reference, data file formats, Profiling Instrument Type, Protocol, depths, study design, cell type, reported ethnicity and library metadata schema version all seem like information we should not try to capture at the dataset level.  Most datasets (especially as they grow larger) will not have consistent info here.  If we wanted to provide genome reference and data file formats, I think we should consider making them calculated values.

scripts/workspace-conversion/README.md
```diff
@@ -0,0 +1,61 @@
...
+| 18 | Data File Formats | `library:dataFileFormats`<br>ex: "TXT, MAF" | `files.0.dcat:mediaType`<br>`files.0.count = 0`<br>`files.0.byteSize = 0`<br>Note: No way of knowing how many files match each file format |
+| 19 | Profiling Instrument Type | `library:technology` |  |
+| 20 | Profiling Protocol | `library:profilingProtocol` |  |
+| 21 | Depth of Sequencing Coverage (Average) | `library:coverage` |  |
+| 22 | Study Design | `library:studyDesign`<br>ex: "Tumor/Normal" | `samples.disease.0`<br>Note: concat value with "Primary Disease Site"? |
+| 23 | Cell Type | `library:cellType`<br>ex: "Primary tumor cell, Whole blood" | `samples.disease.0`<br>Note: concat value with "Primary Disease Site"? |
+| 24 | Reported Ethnicity | `library:ethnicity` |  |
+| 25 | Cohort Country of Origin | `library:cohortCountry`<br>ex: "USA" |  |
+| 26 | Requires External Approval | `library:requiresExternalApproval` |  |
```
kreinold, line 58: This and structured data use limitations should probably be addressed by mapping to Data Use ontology data use modifier terms but it depends upon how they're used in the data.  I'm happy to try to help but Jonathan Lawson, Greg Rushton would be helpful here too.

scripts/workspace-conversion/README.md
lu-c: The two fields make sense separately, but what I meant by "conflicts" is that I think when we map it to the catalog entry format, I was suggesting storing them in the same field - I'm not sure we have two separate places to store the custodian and owner in this current format? Happy to hear suggestions tho!

scripts/workspace-conversion/README.md
lu-c: I think I'm going to leave it in `contributors.institution` AND add it to `prov:wasAssociatedWith`, if that's okay!

scripts/workspace-conversion/README.md
lu-c: Oh ah, yeah, my bad, let me update that - we do have a recognized entry for "General Research Use" too

scripts/workspace-conversion/README.md
lu-c: Ok! We can add it as a column / new field, but with no expectation that anyone else will see or use it :)

scripts/workspace-conversion/README.md
lu-c: we have decided we will calculate it at a later time - right now, we are going to just add the numbers as is

scripts/workspace-conversion/README.md
lu-c: Ok, we will add these to the list of fields that should be calculated in teh future! :)

scripts/workspace-conversion/README.md
lu-c: We will add it as a value in the diseases array, and fix it later after we can examine some more of the generated data

scripts/workspace-conversion/README.md
lu-c: Thank you! We are now storing those in two separate fields, though I believe they are not being displayed in the front end as of right now.

scripts/workspace-conversion/workspace-to-dataset.py
```diff
@@ -0,0 +1,296 @@
...
+        return "TerraCore:NPC"
+    if lowerPolicy == "not-for-profit, non-commercial use ony":
+        return "TerraCore:NPC2"
+    return policyString
+
+
+def mapDataModality(modalityArray):
+    ret = []
+    for modality in modalityArray:
+        if modality == "Epigenomic":
```
pshapiro4broad, line 122: I'm not a python programmer, but can this be done using a dictionary instead of an `if`?

scripts/workspace-conversion/workspace-to-dataset.py
```diff
@@ -0,0 +1,296 @@
...
+import requests
+
+urlRoot = os.environ.get("RAWLS_URL") or "https://rawls.dsde-prod.broadinstitute.org"
+urlWorkspace = f"{urlRoot}/api/workspaces"
+workspaceNamespace = os.environ.get("WORKSPACE_NAMESPACE")
+workspaceName = os.environ.get("WORKSPACE_NAME")
+user = os.environ.get("GCLOUD_USER") or "datacatalogadmin@test.firecloud.org"
+
+
+def logResponse(response, message):
```
pshapiro4broad, line 37: What guidelines do you follow for writing python code? PEP8 is one common convention to follow; regarding function names [it says that they should be lowercase with underscores](https://peps.python.org/pep-0008/#function-and-variable-names).

scripts/workspace-conversion/workspace-to-dataset.py
```diff
@@ -0,0 +1,296 @@
...
+    if 200 <= response.status_code and response.status_code < 300:
```
pshapiro4broad, line 38: I think `ok` is more idiomatic way to write this
```suggestion
    if response.ok:
```

scripts/workspace-conversion/workspace-to-dataset.py
```diff
@@ -0,0 +1,296 @@
...
+#   vault login -method=github token=$(cat ~/.github-token)
+#   vault read secret/dsde/terra/kernel/dev/dev/catalog/tests/userAdmin
+#
+# Run:
+#   `python3 workspace-to-dataset.py`
+#   `GCLOUD_USER={your-email} python3 generate-tdr-snapshot-metadata.py`
+#
+# Environment Variables:
+#   GCLOUD_USER: user account email
+#   CATALOG_SERVICE_URL: url for the catalog service to update the updata in
```
pshapiro4broad, line 23: typo?
```suggestion
#   CATALOG_SERVICE_URL: url for the catalog service to update the data in
```

scripts/workspace-conversion/workspace-to-dataset.py
```diff
@@ -0,0 +1,296 @@
...
+    metadata["dct:title"] = wsAttributes.get("library:datasetName", None)
+    metadata["dct:version"] = wsAttributes.get("library:datasetVersion", None)
+    metadata["dct:description"] = wsAttributes.get("library:datasetDescription", None)
+    metadata["TerraDCAT_ap:hasOwner"] = wsAttributes.get("library:datasetOwner", None)
+    metadata["TerraDCAT_ap:hasDataCollection"] = []
+    if "library:datasetOwner" in wsAttributes:
+        metadata["TerraDCAT_ap:hasDataCollection"].append({})
+        metadata["TerraDCAT_ap:hasDataCollection"][0]["dct:identifier"] = wsAttributes[
+            "library:datasetOwner"
+        ]
```
pshapiro4broad, line 237: Can this dictionary be created as an argument to append? Something like
```suggestion
        metadata["TerraDCAT_ap:hasDataCollection"].append({"dct:identifier": wsAttributes["library:datasetOwner"]})
```

pshapiro4broad (approved):
A few comments. I haven't looked at our python very closely, but I'm wondering if we have coding conventions for python code like we do for Java. I've used [PEP8](https://peps.python.org/pep-0008/) before, and I know [google has a style guide too](https://google.github.io/styleguide/pyguide.html).

scripts/workspace-conversion/workspace-to-dataset.py
lu-c: Haha, i'm just using nodejs standards - I can update it to this lowercase with underscoring style!

scripts/workspace-conversion/workspace-to-dataset.py
lu-c: Ah, updated to this :D

scripts/workspace-conversion/workspace-to-dataset.py
```diff
@@ -0,0 +1,248 @@
...
+#   python3 -m pip install requests
+#
+#   Authorization: Access the password in the vault using the appropriate command:
+#      (you will need this for the auth dialog that the script will pull up)
+#   vault login -method=github token=$(cat ~/.github-token)
+#   vault read secret/dsde/terra/kernel/dev/dev/catalog/tests/userAdmin
+#
+# Run:
+#   `python3 workspace-to-dataset.py`
+#   `export GCLOUD_USER={your-email}; export WORKSPACE_NAMESPACE={workspace-namespace}; export WORKSPACE_NAME={workspace-name}; python3 workspace-to-dataset.py`
```
fboulnois, line 19: Potential future improvements: having command line arguments and validating them?

scripts/workspace-conversion/workspace-to-dataset.py
```diff
@@ -0,0 +1,248 @@
...
+    return list(filter(None, ret))
+
+
+def generate_catalog_metadata(workspace):
+    print("Generating workspace metadata")
+
+    wsAttributes = workspace["workspace"]["attributes"]
+
+    # Set empty up the major objects
+    metadata = {}
```
fboulnois, line 155: Would there be any way to structure the code for this object in a clearer way so it's obvious at a glance what the nesting is for various fields? For example: https://github.com/DataBiosphere/terra-data-catalog/blob/main/scripts/hca/create-hca-collection.py#L175

scripts/workspace-conversion/workspace-to-dataset.py
```diff
@@ -0,0 +1,248 @@
...
+    metadata["prov:wasGeneratedBy"].append(
+        {
+            "TerraCore:hasDataModality": map_data_modality(
+                wsAttributes.get("library:datatype", {}).get("items", [])
+            )
+        }
+    )
+    metadata["files"] = []
+    fileList = wsAttributes.get("library:dataFileFormats", {}).get("items", [])
+    for x in fileList:
```
fboulnois, line 227:
```suggestion
    for ext in fileList:
```

nit: minor suggestion for clarity

scripts/workspace-conversion/workspace-to-dataset.py
```diff
@@ -0,0 +1,248 @@
...
+    "geographical restriction": "TerraCore:GS",
+    "publication moratorium": "TerraCore:MOR",
+    "return to database/resource": "TerraCore:RT",
+    "non commercial use only": "TerraCore:NCU",
+    "not-for-profit use only": "TerraCore:NPC",
+    "not for profit use only": "TerraCore:NPC",
+    "not-for-profit, non-commercial use ony": "TerraCore:NPC2",
+}
+
+dataModalityMap = {
```
fboulnois, line 57: Are there any modalities from here that are relevant? https://github.com/DataBiosphere/terra-data-catalog/blob/main/scripts/hca/create-hca-collection.py#L59

fboulnois (commented):
A few suggestions:

pshapiro4broad (approved):
Still looks good to me. I'd like to see what an imported workspace looks like in the catalog UI, can you do a demo at some point? It will probably require faking out some code in the catalog service locally to get that working, let me know and we can work it together.

scripts/workspace-conversion/workspace-to-dataset.py
lu-c: oh! I'll merge that map in! I got this list of possible modalities from Kathy - And she recommended filtering out anything that didnt match it. Having the other mapping matches the data from the field better!

scripts/workspace-conversion/workspace-to-dataset.py
```diff
@@ -0,0 +1,248 @@
...
+            None,
+            [
+                wsAttributes.get("library:diseaseOntologyLabel", None),
+                wsAttributes.get("library:indication", None),
+                wsAttributes.get("library:primaryDiseaseSite", None),
+                wsAttributes.get("library:studyDesign", None),
+                wsAttributes.get("library:cellType", None),
+            ],
+        )
+    )
```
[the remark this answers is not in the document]
lu-c: I tried breaking up some of the more complicated blocks for this - anything that looked better with an `if` statement. I think the samples block is pretty easy to read tho, so I left some of them as is. What do you think of these changes? :O

scripts/workspace-conversion/workspace-to-dataset.py
lu-c: done!

scripts/workspace-conversion/workspace-to-dataset.py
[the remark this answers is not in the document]
fboulnois: Yep this looks great, thank you!

scripts/workspace-conversion/workspace-to-dataset.py
lu-c: ah definitely would be nice to add

PR #79, opened 2022-05-05
[DC-326] Add new test configs for alpha and staging

s-rubenstein:


integration/src/main/java/scripts/testscripts/SnapshotDatasetOperations.java
```diff
@@ -132,7 +133,7 @@ private void previewUserJourney(CatalogClient client) throws ApiException {
     @SuppressWarnings("unchecked")
     Map<String, String> row = (Map<String, String>) sampleTable.getRows().get(0);
 
-    assertThat(row, hasEntry(is("sample_id"), is("sample5")));
+    assertThat(row, hasEntry(is("sample_id"), is(notNullValue())));
```
s-rubenstein, line 136: This change is because "sample5" exact matching is dependent on the dataset that is created in the environment, not on our test setup. We want to make sure there is some data there, but I don't think we actually need to test _exactly_ what it is, and this makes the test less brittle.

integration/src/main/java/scripts/testscripts/SnapshotDatasetOperations.java
```diff
@@ -188,7 +189,7 @@ private void resultHasDatasetWithRoles(List<Object> datasets) {
         assertThat(dataset, hasEntry(is("id"), is(datasetId.toString())));
         @SuppressWarnings("unchecked")
         List<Object> roles = (List<Object>) dataset.get("roles");
-        assertThat(roles, containsInAnyOrder("steward", "admin"));
+        assertThat(roles, hasSize(not(0)));
```
s-rubenstein, line 192: So, technically, `steward` is the only one we should be confident that exists in the roles. We get `admin` in dev, because the datacatalogadmin account is part of the _DataRepoAdmins_ group (not the _DataCatalogAdmins_) - this is not true in all environments. However, all the contains matchers I could find expect there to be an exact match of number of results (there is no `includes` or non-comprehensive `contains`). Because it differs based on group membership out of our control, I wanted to change it to just test to make sure we are getting _at least one_ role, because that would confirm that we are converting properly.

pshapiro4broad, line 192: Does this work?
```suggestion
        assertThat(roles, not(empty()));
```

Or we could assert the one element we know will be there
```suggestion
        assertThat(roles, hasItem("steward"));
```

integration/src/main/java/scripts/testscripts/SnapshotDatasetOperations.java
pshapiro4broad, line 136: Can you directly assert against `notNullValue()`?
```suggestion
    assertThat(row, hasEntry(is("sample_id"), notNullValue()));
```

integration/src/main/java/scripts/testscripts/SnapshotDatasetOperations.java
pshapiro4broad: Unlike the preview, the content of the roles should be asserted since our code depends on it not changing over time.

pshapiro4broad (approved):
Looks good, just a few comments. If TC's change merges up before this, you'll have to update the test suites to include the new test configuration from that PR.

integration/src/main/java/scripts/testscripts/SnapshotDatasetOperations.java
s-rubenstein: So, in that case, we need to make sure we are either not in, or are in, the DataRepoAdmins group. I assume we should talk to Jade Core to get our admin account added to that group in all envs?

s-rubenstein: … nevermind, that's not actually blocking. How did I miss that hasItem isn't a complete equivalence.

integration/src/main/java/scripts/testscripts/SnapshotDatasetOperations.java
```diff
@@ -188,7 +189,7 @@ private void resultHasDatasetWithRoles(List<Object> datasets) {
         assertThat(dataset, hasEntry(is("id"), is(datasetId.toString())));
         @SuppressWarnings("unchecked")
         List<Object> roles = (List<Object>) dataset.get("roles");
-        assertThat(roles, containsInAnyOrder("steward", "admin"));
+        assertThat(roles, hasItem("steward"));
```
fboulnois, line 192: I saw the earlier comments but I still don't think I full understand this change. Can you summarize why it's necessary again?

fboulnois (approved):
I have one question but otherwise lgtm:

integration/src/main/java/scripts/testscripts/SnapshotDatasetOperations.java
s-rubenstein: So, the "admin" role is granted via having a permission on an admin group on DataRepo - completely tangent to the actual behavior of the Data Catalog, and isn't needed for Data Catalog operations, so we shouldn't test on its existence.

PR #80, opened 2022-05-06
DC-333: Add configuration for Rawls status check

fboulnois:


service/src/main/java/bio/terra/catalog/config/RawlsConfiguration.java
```diff
@@ -0,0 +1,6 @@
+package bio.terra.catalog.config;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+@ConfigurationProperties(prefix = "catalog.rawls")
+public record RawlsConfiguration(String basePath, String resourceId) {}
```
pshapiro4broad, line 6: I don't think we need `resourceId` for this config

service/src/main/java/bio/terra/catalog/rawls/RawlsService.java
```diff
@@ -0,0 +1,34 @@
...
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+@Component
+public record RawlsService(RawlsConfiguration rawlsConfig) {
+  private static final Logger logger = LoggerFactory.getLogger(RawlsService.class);
+
+  @Autowired
+  public RawlsService {}
```
pshapiro4broad, line 15: Is this necessary? It's good to see a `record` `Component`, I'd been meaning to try it out.

service/src/main/java/bio/terra/catalog/config/RawlsConfiguration.java
fboulnois: Whoops, a bit of copy-paste from the other configurations.

service/src/main/java/bio/terra/catalog/rawls/RawlsService.java
fboulnois: Actually the `record Component` prevents `mockito` from using `spy`, so I had to switch it back to a regular component.

rawls-client/swagger.gradle
```diff
@@ -0,0 +1,50 @@
...
+    implementation 'org.glassfish.jersey.media:jersey-media-multipart'
+
+    implementation 'io.swagger.core.v3:swagger-annotations'
+    swaggerCodegen 'io.swagger.codegen.v3:swagger-codegen-cli'
+}
+
+def artifactGroup = "${group}.rawls"
+
+task downloadRawlsYaml(type: Download) {
+    // where to download the Cloud Profiler agent https://cloud.google.com/profiler/docs/profiling-java
```
pshapiro4broad, line 15: I forgot to delete this comment
```suggestion
```

service/src/test/java/bio/terra/catalog/rawls/RawlsServiceTest.java
```diff
@@ -0,0 +1,62 @@
...
+
+  @BeforeEach
+  void beforeEach() {
+    rawlsService = spy(rawlsServiceReal);
+  }
+
+  @Test
+  void status() throws Exception {
+    SystemStatusSystems status = new SystemStatusSystems().ok(true);
+    when(rawlsService.status()).thenReturn(status);
```
pshapiro4broad, line 36: It's not a big deal but to make this test structured like the other service tests (`SamServiceTest` and `DatarepoServiceTest`), it would be more consistent to mock the method that creates the API, instead of a higher level method. In this case you'd mock a new method like `statusApi()`.

The main benefit of writing a test this way is you get coverage of the code that immediately surrounds the service's API layer. There's not much code for `status` but there will be more for other Rawls operations.

pshapiro4broad (approved):
Looks good, just one comment about the mocking in the test code.

pshapiro4broad:
Also, to fix the sonar warning about `registerStatusCheck`, I'd remove the `VisibleForTesting` annotation from it for now. The warning is generated because the visibility is not only needed for testing, but for the code in the app to compile as well.

service/src/main/java/bio/terra/catalog/rawls/RawlsService.java
pshapiro4broad: That makes sense. Looking at the code again, if we moved the code that created the APIs into the configuration class (and renamed it accordingly), and then mocked that instead, that might be a better approach overall. But that can wait for another PR.

service/src/test/java/bio/terra/catalog/rawls/RawlsServiceTest.java
fboulnois: The problem is that since `StatusApi().systemStatus()` returns `void`, I get an error if I try to mock this call. I'm not sure if there's a way to solve this.

service/src/main/java/bio/terra/catalog/rawls/RawlsService.java
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  public boolean statusIsOk() throws ApiException {
+    new StatusApi(getApiClient()).systemStatus();
+    return true;
+  }
+
+  public SystemStatusSystems status() {
+    var result = new SystemStatusSystems();
+    try {
```
petesantos, line 39: I am confused by this try/catch block, but it is probably just my Java understanding.

If the result is not ok, or it is thrown we log the error and add the message to the result - when is the error thrown vs gettting a false on `statusIsOk`?

Could this be simplified?

petesantos (approved):

service/src/main/java/bio/terra/catalog/rawls/RawlsService.java
fboulnois: There are some issues with the Rawls status endpoint that may need to be resolved in the future. As a result, although the Rawls method currently returns nothing or throws, it is possible that `statusIsOk` returns true, false, or throws in the future. That's why I have a condition that checks the result and that also check if the method throws.

service/src/test/java/bio/terra/catalog/rawls/RawlsServiceTest.java
```diff
@@ -0,0 +1,50 @@
...
+  private RawlsService rawlsService;
+
+  @BeforeEach
+  void beforeEach() {
+    rawlsService = spy(rawlsServiceReal);
+    doReturn(statusApi).when(rawlsService).statusApi();
+  }
+
+  @Test
+  void status() throws Exception {
```
pshapiro4broad, line 39: Are the `throws Exception`s still needed? I guess it doesn't hurt to keep them.

pshapiro4broad (approved):
LGTM

PR #82, opened 2022-05-11
DC-423: Remove `spy()` from our service tests

pshapiro4broad:
This change removes the need to use `spy()` in our service tests. Writing tests with `spy()` can be confusing and it's easy to make a mistake. This change also allows us to get 100% code coverage on these classes.

The main change is to move the HTTP API client construction code into a separate class. This lets us mock directly at the API client layer when testing a service class.

WIth this change, all uses of `spy()` have been removed from the code.

While doing this, I also changed the service tests to no longer use `ContextConfiguration` to set up the spring context. Instead, `MockitoExtension` is used to initialize mocks. For these tests, initializing Spring wasn't necessary. This change also enables strict stubbing by default, which is also an improvement IMO. See http://blog.mockito.org/2017/01/clean-tests-produce-clean-code-strict.html.


service/src/main/java/bio/terra/catalog/datarepo/DatarepoClient.java
```diff
@@ -0,0 +1,41 @@
...
+  private final DatarepoConfiguration datarepoConfig;
+  private final Client commonHttpClient;
+
+  @Autowired
+  public DatarepoClient(DatarepoConfiguration datarepoConfig) {
+    this.datarepoConfig = datarepoConfig;
+    this.commonHttpClient = new ApiClient().getHttpClient();
+  }
+
+  public SnapshotsApi snapshotsApi(AuthenticatedUserRequest user) {
```
tcjiang9, line 23: I think we should use the SnapshotsSyncApi I added when possible

service/src/main/java/bio/terra/catalog/datarepo/DatarepoClient.java
```diff
@@ -0,0 +1,41 @@
...
+    return new SnapshotsApi(getApiClient(user));
+  }
+
+  private ApiClient getApiClient(AuthenticatedUserRequest user) {
+    ApiClient apiClient = getApiClient();
+    apiClient.setAccessToken(user.getToken());
+    return apiClient;
+  }
+
+  private ApiClient getApiClient() {
```
tcjiang9, line 33: why have a separate function here?

service/src/test/java/bio/terra/catalog/datarepo/DatarepoClientTest.java
```diff
@@ -0,0 +1,43 @@
...
+            .setToken(TOKEN)
+            .build();
+    var snapshotsClient = client.snapshotsApi(user).getApiClient();
+    assertThat(snapshotsClient.getBasePath(), is(BASE_PATH));
+
+    OAuth oauth = (OAuth) snapshotsClient.getAuthentication("googleoauth");
+    assertNotNull(oauth);
+    assertThat(oauth.getAccessToken(), is(TOKEN));
+
+    var unauthClient = client.unauthenticatedApi().getApiClient();
```
tcjiang9, line 38: should we also assert that the unauth client doesn't have an access token?

service/src/test/java/bio/terra/catalog/datarepo/DatarepoClientTest.java
```diff
@@ -0,0 +1,43 @@
...
+
+  @Test
+  void testApis() {
+    var user =
+        new AuthenticatedUserRequest.Builder()
+            .setEmail("")
+            .setSubjectId("")
+            .setToken(TOKEN)
+            .build();
+    var snapshotsClient = client.snapshotsApi(user).getApiClient();
```
tcjiang9, line 31: should we just expose getApiClient() so we don't have to create a snapshotsApi to then get its ApiClient?

pshapiro4broad: This particular form is only used in test code, as a (very artificial) way to test the created APIs. Since it's not used in non-test code, I don't think we should add an API in `client` for this.

service/src/test/java/bio/terra/catalog/datarepo/DatarepoClientTest.java
pshapiro4broad: The bug turned out to be in the test, not in the service code. It's fixed now.

service/src/main/java/bio/terra/catalog/datarepo/DatarepoService.java
```diff
@@ -108,25 +102,12 @@ public DatasetAccessLevel getRole(AuthenticatedUserRequest user, String snapshot
     }
   }
 
-  @VisibleForTesting
-  SnapshotsApi snapshotsApi(AuthenticatedUserRequest user) {
-    return new SnapshotsApi(getApiClient(user));
+  private SnapshotsApi snapshotsApi(AuthenticatedUserRequest user) {
```
s-rubenstein, line 105: My inclination is to one-line these functions, just given how simple they are. But I am also perfectly fine with them as they are. It looks like we also do that below in SamService, so we should at least keep the approach the same.

service/src/main/java/bio/terra/catalog/rawls/RawlsService.java
```diff
@@ -84,8 +65,8 @@ public DatasetAccessLevel getRole(AuthenticatedUserRequest user, String workspac
     }
   }
 
-  WorkspacesApi workspacesApi(AuthenticatedUserRequest user) {
-    return new WorkspacesApi(getApiClient(user));
+  private WorkspacesApi workspacesApi(AuthenticatedUserRequest user) {
```
s-rubenstein, line 68: Same as the comment in DatarepoService. My inclination is to inline this, but I am okay with this approach. However, we should make sure all our services are taking the same approach.

service/src/main/java/bio/terra/catalog/datarepo/DatarepoService.java
pshapiro4broad: I had left it this way to make it easier to merge in changes from `main` since this was a long running branch. Let me try inlining and see how it looks.

s-rubenstein (approved):
LGTM

service/src/main/java/bio/terra/catalog/datarepo/DatarepoClient.java
```diff
@@ -0,0 +1,38 @@
...
+    // Share one api client across requests.
+    return new ApiClient().setHttpClient(commonHttpClient).setBasePath(datarepoConfig.basePath());
+  }
+
```
tcjiang9, line 30: why get rid of the @visiblefortesting tag?

service/src/test/java/bio/terra/catalog/datarepo/DatarepoClientTest.java
```diff
@@ -0,0 +1,43 @@
...
+class DatarepoClientTest {
+
+  private static final String BASE_PATH = "basepath";
+  private static final String TOKEN = "token";
+  private static final String AUTH_NAME = "googleoauth";
+
+  private final DatarepoClient client = new DatarepoClient(new DatarepoConfiguration(BASE_PATH));
+
+  @Test
+  void testApis() {
```
tcjiang9, line 21: would something like testApiTokens be more informative? This doesn't actually test any APIs

tcjiang9 (approved):

tcjiang9:
Small comments but LGTM

service/src/main/java/bio/terra/catalog/datarepo/DatarepoClient.java
pshapiro4broad: It's only needed if the access is changed due to the need to access a member from test code. We used to need it to make some APIs visible that we would've made private, because there wasn't a way to mock the point where an API was created. Now that this is done in a separate class, the access would no longer be private so the annotation isn't needed.

service/src/test/java/bio/terra/catalog/datarepo/DatarepoClientTest.java
pshapiro4broad: It tests the API objects that the client code produces. The way it tests the objects is by looking at the auth tokens and the base path, but that testing method could change in the future, so it seems better to describe what it's doing rather than how it's doing it.

It could be written as a different test for each API method, but I wanted to add a test that verified that the http client was actually being shared, so it seemed easier to test all APIs in a single method.

PR #83, opened 2022-05-12
DC-332 add customization to our nightly tests

tcjiang9:
Modify [Nightly perf](https://github.com/DataBiosphere/terra-data-catalog/blob/main/.github/workflows/nightly-tests.yml) GitHub Action to support customization of environments and test suites through environment variables.


fboulnois (approved):
👍

.github/workflows/nightly-tests.yml
```diff
@@ -1,10 +1,24 @@
 name: Nightly Tests
 
 on:
-  workflow_dispatch: {}
+  workflow_dispatch:
+    inputs:
+      stage:
```
pshapiro4broad, line 6: I would expect this input to be called `environment`, not `stage`.

integration/src/main/resources/suites/integration.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "FullIntegration",
+  "name": "integration",
```
pshapiro4broad, line 2: This renaming doesn't follow the convention used by other repos. Looking at workspace manager, a their approach is to put the test inside a directory, and use the directory name to determine which environment the test is written for. See https://github.com/DataBiosphere/terra-workspace-manager/tree/main/integration/src/main/resources/suites

Also should this be `dev` not `integration`? Or maybe `pull-request`?

pshapiro4broad (changes requested):
A few minor comments to address. Also, you'll need to update the README to match the new test locations.

Also, could you update the PR so it has a description? If the ticket description is good, you could copy that over.

fboulnois (approved):

.github/workflows/nightly-tests.yml
```diff
@@ -1,10 +1,24 @@
 name: Nightly Tests
 
 on:
-  workflow_dispatch: {}
+  workflow_dispatch:
+    inputs:
+      environment:
+        type: choice
+        description: 'stage to run test'
```
pshapiro4broad, line 8: Update description to match argument name
```suggestion
        description: 'environment to run test in'
```

.github/workflows/nightly-tests.yml
```diff
@@ -97,7 +111,7 @@ jobs:
   notify-de-slack:
     needs: [ build, jib, test-runner ]
     runs-on: ubuntu-latest
-    if: failure()
+    if: ${{ github.event_name != 'workflow_dispatch' && failure() }}
```
pshapiro4broad, line 114: Is this logic correct? If this workflow is triggered by a Jenkins job, would we want to see failures in our slack? If not, how would we be notified of failures? Maybe sending a message directly from Jenkins would be better, if that can be done.

pshapiro4broad (changes requested):
A few more questions / suggestions. Also, the README still needs to be updated.

.github/workflows/nightly-tests.yml
tcjiang9: workflow dispatch is the manual trigger that is done via the bash script I added, so this won't be reported if it's manual trigger and fails

pshapiro4broad: > workflow dispatch is the manual trigger that is done via the bash script I added, so this won't be reported if it's manual trigger and fails

So to be clear; if this is run by a workflow dispatch (which is what Jenkins will use) then errors _won't_ be reported to slack? Why wouldn't we want to see errors caused by the Jenkins job trigger? Or are those errors reported to our slack via some other system?

tcjiang9: this was just done with the script in mind for locally testing the github action, if it will conflict with jenkins then I can change it back

pshapiro4broad: OK; I was asking because I don't know how errors are reported from the Jenkins jobs that run alpha and staging tests. I assume that they don't report status in our slack currently, and it might be easiest to report it in our workflow rather than customizing the Jenkins jobs.

pshapiro4broad (approved):
Looks good

lu-c (approved):
LGTM

PR #85, opened 2022-05-17
Add roles as enum to data repo spec

s-rubenstein:


service/src/test/java/bio/terra/catalog/datarepo/DatarepoServiceTest.java
```diff
@@ -54,13 +55,22 @@ void beforeEach() {
     doReturn(unauthenticatedApi).when(datarepoService).unauthenticatedApi();
   }
 
+  @Test
+  void accessMapping() throws Exception {
+    assertThat(DatarepoService.SNAPSHOT_ROLE_TO_DATASET_ACCESS_LEVEL.get(DatarepoService.STEWARD_ROLE_NAME), is(DatasetAccessLevel.OWNER));
```
pshapiro4broad, line 60: I don't think we should have tests that only test static data structures, as they duplicate the state of the data in test form only. Instead we should have a test on an operation that depends on this object.

common/openapi.yml
```diff
@@ -287,6 +287,17 @@ components:
             type: object
       description: List of catalog datasets
 
+    DatasetAccessLevel:
+      type: string
+      description: The level of access a user or group has on a Dataset. Owners can create and delete
+        dataset entries, writers can modify only, readers can read the data, and discoverers can see
```
pshapiro4broad, line 293: I don't think the catalog needs to distinguish between `writer` and `owner`. Can you describe a case where we'd want to allow a user to modify the contents of a catalog entry but not create or delete it? If a user can overwrite all elements in an entry, they could replace it with an empty one, which would have the same effect as deletion.

My thinking is that we'd want to map workspace writers to the `reader` role, as we'd want to limit catalog entry creation to the owner of the data, not a user who has write access to the workspace.

service/src/main/java/bio/terra/catalog/datarepo/DatarepoService.java
```diff
@@ -47,11 +55,18 @@ public DatarepoService(DatarepoConfiguration datarepoConfig) {
...
-      return snapshotsApi(user)
+      Map<String, List<String>> response = snapshotsApi(user)
           .enumerateSnapshots(null, MAX_DATASETS, null, null, null, null, null)
           .getRoleMap();
+      Map<String, List<DatasetAccessLevel>> idsToDatasetAccessLevelList = new java.util.HashMap<>(Map.of());
+      idsToDatasetAccessLevelList.putAll(
+          response.entrySet().stream().collect(Collectors.toMap(
+              Map.Entry::getKey,
+              entry -> entry.getValue().stream().map(SNAPSHOT_ROLE_TO_DATASET_ACCESS_LEVEL::get).toList())));
+      return idsToDatasetAccessLevelList;
```
pshapiro4broad, line 69: Is this code necessary? I would expect that you could return the result of the streaming operation directly
```suggestion
      return  response.entrySet().stream().collect(Collectors.toMap(
              Map.Entry::getKey,
              entry -> entry.getValue().stream().map(SNAPSHOT_ROLE_TO_DATASET_ACCESS_LEVEL::get).toList());;
```

service/src/main/java/bio/terra/catalog/datarepo/DatarepoService.java
```diff
@@ -47,11 +55,18 @@ public DatarepoService(DatarepoConfiguration datarepoConfig) {
     this.commonHttpClient = new ApiClient().getHttpClient();
   }
 
-  public Map<String, List<String>> getSnapshotIdsAndRoles(AuthenticatedUserRequest user) {
+
+  public Map<String, List<DatasetAccessLevel>> getSnapshotIdsAndRoles(AuthenticatedUserRequest user) {
```
pshapiro4broad, line 59: If we're going to create an API for roles, it might be more streamlined to define them as a hierarchy (discoverer -> reader -> owner) where `owner` can do everything `reader` can do, and `reader` can do everything `discoverer` can do. Then this could return `Map<String, DatasetAccessLevel>`. If we wanted to support a more complex mix of behaviors, we should use actions / capabilities instead.

pshapiro4broad (approved):
I like this idea, but I think we should hold off on adding `writer` until we have a use case that requires it.

service/src/main/java/bio/terra/catalog/datarepo/DatarepoService.java
s-rubenstein: Hmmm… I agree that makes things easier for permissions checks in the future. I am wondering how to pull that off with the swagger generated enums _and_ backwards compatibility/returning string values.

I think we could do a bigger refactor if we were okay breaking the Catalog UI for a bit while this released, but I think to avoid breaking, we either need to introduce some complexity in the UI, version the endpoints and create a new version (significant back end complexity), accept the subpar mapping, or create a custom class and not use the openAPI generation for this enum.

I lean towards the last one (custom class, no openAPI generation), but am curious your thoughts.

common/openapi.yml
```diff
@@ -303,6 +303,7 @@ components:
       description: |
         ⚠️ Do not add sensitive data. All catalog entries are publicly accessible to all Terra users. ⚠️
       type: object
+      required: [storageSystem, storageSourceId]
```
s-rubenstein, line 306: It seemed like this was an oversight, so I just wanted to driveby fix it. We had a test that was generating NPE for using nulls here, and I couldn't figure out any realistic case someone would want to create a catalog entry without a storage system or storage source id.

s-rubenstein:
One thing that is weird is that we are going to return a list of Data Catalog roles rn in the form of a role array. I intend to fix this, but that one would be the breaking change, so I am going to spin it out _again_ and do that as a follow on PR to make that change as small as possible.

service/src/main/java/bio/terra/catalog/datarepo/DatarepoService.java
```diff
@@ -30,9 +31,12 @@ public class DatarepoService {
   public static final String READER_ROLE_NAME = "reader";
   public static final String DISCOVERER_ROLE_NAME = "discoverer";
 
-  private static final List<String> OWNER_ROLES = List.of(ADMIN_ROLE_NAME, STEWARD_ROLE_NAME);
-  private static final List<String> READER_ROLES =
-      List.of(ADMIN_ROLE_NAME, STEWARD_ROLE_NAME, READER_ROLE_NAME, DISCOVERER_ROLE_NAME);
+  public static final Map<String, DatasetAccessLevel> SNAPSHOT_ROLE_TO_DATASET_ACCESS_LEVEL =
```
pshapiro4broad, line 34: Can this be `private`? Also I prefer shorter names when meaning aren't ambiguous. Since we only deal with snapshot roles here I think `ROLE_TO_DATASET_ACCESS_LEVEL` or even `ROLE_TO_DATASET_ACCESS` would be fine.

service/src/main/java/bio/terra/catalog/datarepo/DatarepoService.java
```diff
@@ -47,11 +51,21 @@ public DatarepoService(DatarepoConfiguration datarepoConfig) {
     this.commonHttpClient = new ApiClient().getHttpClient();
   }
 
-  public Map<String, List<String>> getSnapshotIdsAndRoles(AuthenticatedUserRequest user) {
+  public Map<String, List<DatasetAccessLevel>> getSnapshotIdsAndRoles(
```
pshapiro4broad, line 54: Do we still need `List` here?
```suggestion
  public Map<String, DatasetAccessLeve>> getSnapshotIdsAndRoles(
```

service/src/main/java/bio/terra/catalog/datarepo/DatarepoService.java
```diff
@@ -76,18 +90,21 @@ public SnapshotPreviewModel getPreviewTable(
...
-      var roles = rolesForAction(action);
-      return snapshotsApi(user).retrieveUserSnapshotRoles(id).stream().anyMatch(roles::contains);
+      List<String> roles = snapshotsApi(user).retrieveUserSnapshotRoles(id);
+      for (DatasetAccessLevel datasetAccessLevel : DatasetAccessLevel.values()) {
+        if (roles.stream()
+            .map(SNAPSHOT_ROLE_TO_DATASET_ACCESS_LEVEL::get)
+            .anyMatch(
+                roleAsDatasetAccessLevel -> roleAsDatasetAccessLevel.equals(datasetAccessLevel))) {
+          return datasetAccessLevel;
+        }
```
pshapiro4broad, line 103: If this was a function, of TDR roles (List<String>) to catalog role (DatasetAccessLevel), I think you could use it in `getSnapshotIdsAndRoles()` above and not duplicate the logic.

service/src/main/java/bio/terra/catalog/service/DatasetService.java
```diff
@@ -72,8 +73,8 @@ private void collectDatarepoDatasets(
     // Merge the permission (role) data into the metadata results.
     for (Dataset dataset : datasets) {
       ArrayNode roles = objectMapper.createArrayNode();
-      for (String role : roleMap.get(dataset.storageSourceId())) {
-        roles.add(TextNode.valueOf(role));
+      for (DatasetAccessLevel role : roleMap.get(dataset.storageSourceId())) {
+        roles.add(TextNode.valueOf(role.toString()));
```
pshapiro4broad, line 77: Very minor but I usually prefer `String.valueOf()` as it works on all java types, unlike `toString()` which only works on Objects.
```suggestion
        roles.add(TextNode.valueOf(String.valueOf(role)));
```

service/src/main/java/bio/terra/catalog/service/dataset/DatasetAccessLevel.java
```diff
@@ -0,0 +1,37 @@
...
+  @Override
+  public String toString() {
+    return name;
+  }
+
+  public boolean hasAction(SamAction action) {
+    return this.allowedActions.contains(action);
+  }
+
+  public static DatasetAccessLevel fromValue(String input) {
```
pshapiro4broad, line 29: This appears to be unused; if so, I would remove it.

service/src/main/java/bio/terra/catalog/service/dataset/DatasetAccessLevel.java
```diff
@@ -0,0 +1,37 @@
...
+import bio.terra.catalog.iam.SamAction;
+import java.util.EnumSet;
+import java.util.Set;
+
+public enum DatasetAccessLevel {
+  OWNER("owner", EnumSet.allOf(SamAction.class)),
+  READER("reader", EnumSet.of(SamAction.READ_ANY_METADATA)),
+  DISCOVERER("discoverer", Set.of());
+
+  private final String name;
```
pshapiro4broad, line 12: Might be worth commenting that the UI depends on this field's contents and changing it will require changing the UI code.

service/src/main/java/bio/terra/catalog/service/dataset/DatasetAccessLevel.java
```diff
@@ -0,0 +1,37 @@
+package bio.terra.catalog.service.dataset;
+
+import bio.terra.catalog.iam.SamAction;
+import java.util.EnumSet;
+import java.util.Set;
+
+public enum DatasetAccessLevel {
+  OWNER("owner", EnumSet.allOf(SamAction.class)),
+  READER("reader", EnumSet.of(SamAction.READ_ANY_METADATA)),
+  DISCOVERER("discoverer", Set.of());
```
pshapiro4broad, line 10: I'm not sure this is correct (although it is what the current code is doing). I think `DISCOVERER` can also `READ_ANY_METADATA` (given the way this action is used used in our code, which doesn't quite match the action name).

If we did a permission check for `export` or `preview` operations, then we'd need another action for those operations. Right now those permission checks are done in the storage system services.

pshapiro4broad (approved):
Looks good, I have a few comments

service/src/main/java/bio/terra/catalog/service/dataset/DatasetAccessLevel.java
```diff
@@ -0,0 +1,29 @@
...
+import bio.terra.catalog.iam.SamAction;
+import java.util.EnumSet;
+import java.util.Set;
+
+public enum DatasetAccessLevel {
+  OWNER("owner", EnumSet.allOf(SamAction.class)),
+  READER("reader", EnumSet.of(SamAction.READ_ANY_METADATA)),
+  DISCOVERER("discoverer", EnumSet.of(SamAction.READ_ANY_METADATA));
+
+  // The UUI depends on this fields contents. Changing these will require a change to the UI.
```
fboulnois, line 12:
```suggestion
  // The UI depends on this fields contents. Changing these will require a change to the UI.
```

micro-nit: misspelled?

fboulnois (approved):
lgtm!

service/src/main/java/bio/terra/catalog/datarepo/DatarepoService.java
pshapiro4broad: After seeing your comment above:

> One thing that is weird is that we are going to return a list of Data Catalog roles rn in the form of a role array. I intend to fix this, but that one would be the breaking change, so I am going to spin it out again and do that as a follow on PR to make that change as small as possible.

This change could be done later too. And/or do the mapping from X -> List<X> in the caller instead of in the services.

pshapiro4broad:
FYI `jib` will fail in `trivy` until the container is fixed. If the only build failure is the known trivy failure, I think it's OK to merge with failures.

s-rubenstein:
@pshapiro4broad looks like I don't have permissions to merge this PR if there are failing tests. So it may need to wait unless you can hit the button.

service/src/test/java/bio/terra/catalog/datarepo/DatarepoServiceTest.java
```diff
@@ -71,29 +72,27 @@ void getSnapshotsException() throws Exception {
   }
 
   @Test
-  void userHasActionReader() throws Exception {
+  void getRoleReader() throws Exception {
     var id = UUID.randomUUID();
     when(snapshotsApi.retrieveUserSnapshotRoles(id))
         .thenReturn(List.of(DatarepoService.READER_ROLE_NAME));
-    assertTrue(datarepoService.userHasAction(user, id.toString(), SamAction.READ_ANY_METADATA));
+    assertEquals(datarepoService.getRole(user, id.toString()), DatasetAccessLevel.READER);
```
pshapiro4broad, line 79: I would use `assertThat()` here instead, since it's usually better at printing out differences between objects. If you do use `assertEquals()`, you should swap the argument order, since the expected value comes first.
```suggestion
    assertThat(datarepoService.getRole(user, id.toString()), is(DatasetAccessLevel.READER));
```

PR #88, opened 2022-05-23
DC-364: Revert "Ensure nightly scheduled test has a default env (#86)"

fboulnois:
This reverts commit b5a479bfb7dd2c23150cf8b553cefec688ed742f.

The GitHub community recommendation will actually break the action. I switched to another solution proposed by @pshapiro4broad .

Also includes the fix from: https://github.com/DataBiosphere/terra-data-catalog/pull/87

### DIFF
--- a/scripts/manually-run-nightly-tests.sh
+++ b/scripts/manually-run-nightly-tests.sh
@@ -2,7 +2,7 @@
 
 set -eu
 
-TOKEN=$(cat "$HOME/.github-token")
+TOKEN=$(find "$HOME" -type f -maxdepth 2 -exec grep -l "^ghp_" {} \; -quit | xargs cat)
 TEST_ENV="perf"
 
 # ensure the script always runs in the parent directory terra-data-catalog
--- a/scripts/manually-run-nightly-tests.sh
+++ b/scripts/manually-run-nightly-tests.sh
@@ -1,0 +1,29 @@
+#!/bin/sh
+
+TOKEN=$(cat $HOME/.github-token)
+TEST_ENV="perf"
+
+# ensure the script always runs in the parent directory terra-data-catalog
+CWD=$(dirname $(dirname $(readlink -f "$0")))
+cd $CWD
+
+# get the helm chart versions for the perf env
+curl -H "Authorization: token ${TOKEN}" \
+    -H 'Accept: application/vnd.github.v3.raw' \
+    -L https://api.github.com/repos/broadinstitute/terra-helmfile/contents/versions/app/dev.yaml \
+    --create-dirs -o "integration/src/main/resources/rendered/dev.yaml"
+curl -H "Authorization: token ${TOKEN}" \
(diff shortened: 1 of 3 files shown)


pshapiro4broad (commented):
I see that there's a default specified for `environment`
```
        default: 'perf'
```
does that not work? Or maybe the trigger is overriding the default with an empty string somehow?

fboulnois:
> I see that there's a default specified for `environment`
> 
> ```
>         default: 'perf'
> ```
> 
> does that not work? Or maybe the trigger is overriding the default with an empty string somehow?

The problem is that only works when doing a `workflow_dispatch`, but it is empty when running a cron job.

scripts/manually-run-nightly-tests.sh
```diff
@@ -0,0 +1,31 @@
...
+# render the service account secrets
+vault read secret/dsde/firecloud/dev/common/firecloud-sa -format=json | \
+    jq -r .data.key | base64 -d > "integration/src/main/resources/rendered/user-delegated-sa.json"
+
+vault read secret/dsde/terra/kernel/perf/common/testrunner/testrunner-sa -format=json | \
+    jq -r .data.key | base64 -d > "integration/src/main/resources/rendered/testrunner-perf.json"
+
+# run the perf test suite
+./gradlew --build-cache runTest --args="suites/$TEST_ENV/FullIntegration.json build/reports"
+./gradlew --build-cache uploadResults --args="CompressDirectoryToTerraKernelK8S.json build/reports"
```
pshapiro4broad, line 31: Is `uploadResults` necessary for a manual test run? I'm not sure what the purpose of this script is; is it to test out the code in the GitHub workflow? Since it duplicates the code in the workflow, it may not be a great way to do that, as now we'll have to change things in both places to keep them in sync.

pshapiro4broad (approved):
Looks OK but I'm wary about duplicating the GitHub workflow code into a shell script.

I guess it could be refactored in a way so that both the workflow and the script use the same code, and environment variables are used to pass the three secret values to the script.

lu-c (approved):
looks good to me!

scripts/manually-run-nightly-tests.sh
fboulnois: As the GitHub Action has not been working since Friday afternoon, Gary recommended I run the steps manually. Instead of manually running each line in the GitHub Action I thought it would be more consistent running it in a shell script which duplicates the functionality. I agree with you generally we should avoid duplicating code, but this was more of a tool I created for myself to be able to run the steps without having to manually run each line. I thought I would share it with the team in case this happens again in the future, but I can remove it if you don't think it is useful. I think rewriting our action to use the shell script would require a lot more effort.

pshapiro4broad: Ok. I think it's OK to keep it in for now, although I'd rather remove it than maintain it if it's not something we need to use often.

PR #89, opened 2022-05-23
DC-361: Dynamically create TDR dataset once for all integration tests

pshapiro4broad:
Instead of creating TDR snapshots using a single test dataset, create a new dataset for every test suite run. To avoid the overhead of creating and deleting a dataset for each test, a singleton object is used to store the test dataset once created. A JVM shutdown hook is used to clean up this dataset.

This is slightly slower than the current code, but should fix issues we've seen where overlapping test runs fail due to dataset locking issues.

With this change, the integration tests take ~7 minutes, where before they took ~5 minutes.


integration/src/main/java/scripts/api/TdrDatasetsApi.java
```diff
@@ -0,0 +1,166 @@
...
+import bio.terra.datarepo.model.TableModel;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import scripts.client.DatarepoClient;
+
+public class TdrDatasetsApi {
```
tcjiang9, line 21: why not extend datasetsapi?

integration/src/main/java/scripts/api/SnapshotsApi.java
```diff
@@ -0,0 +1,58 @@
...
+import bio.terra.datarepo.model.DatasetModel;
+import bio.terra.datarepo.model.PolicyMemberRequest;
+import bio.terra.datarepo.model.SnapshotRequestContentsModel;
+import bio.terra.datarepo.model.SnapshotRequestModel;
+import java.util.UUID;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import scripts.client.DatarepoClient;
+
+public class SnapshotsApi {
```
tcjiang9, line 13: why not extend snapshotsapi anymore?

integration/src/main/java/scripts/api/TdrDatasetsApi.java
```diff
@@ -0,0 +1,166 @@
...
+                        new Column("biological_sex", TableDataType.STRING),
+                        new Column("age", TableDataType.INTEGER)),
+                    new Table(
+                        "sample",
+                        new Column("sample_id", TableDataType.STRING),
+                        new Column("participant_id", TableDataType.STRING),
+                        new Column("files", TableDataType.STRING).arrayOf(true),
+                        new Column("type", TableDataType.STRING))));
+    return new DatasetRequestModel()
+        .name("catalog_integration_test_" + System.currentTimeMillis())
```
tcjiang9, line 100: if we're trying to guard against locking problems related to multiple tests running at the same time, is there a way to incorporate a uuid in here instead of the system time?

integration/src/main/java/scripts/api/TdrDatasetsApi.java
```diff
@@ -0,0 +1,166 @@
...
+                      deleteDataset(testDataset.getId());
+                      testDataset = null;
+                    } catch (Exception e) {
+                      log.error("error deleting dataset during shutdown", e);
+                    }
+                  }
+                }));
+  }
+
+  public static synchronized TdrDatasetsApi createApi(DatarepoClient apiClient) {
```
tcjiang9, line 49: why does this also need to be a singleton?

integration/src/main/java/scripts/api/TdrDatasetsApi.java
pshapiro4broad: Because I want to limit the API exposed to the caller. I don't want to allow a caller to call `createDataset()` or `deleteDataset()` directly, for example.

integration/src/main/java/scripts/api/TdrDatasetsApi.java
```diff
@@ -0,0 +1,166 @@
...
+import java.util.UUID;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import scripts.client.DatarepoClient;
+
+public class TdrDatasetsApi {
+  private static final Logger log = LoggerFactory.getLogger(TdrDatasetsApi.class);
+  private static final String TEST_BILLING_PROFILE_PREFIX = "catalog_test_profile";
+  private static TdrDatasetsApi theApi;
+  private DatasetModel testDataset;
```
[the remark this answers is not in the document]
pshapiro4broad: I think making the class a singleton makes it easer to reason about how it will work, but that's a pretty subjective measure. In this case though, `datasetsApi` must also be static, since it's needed to delete the dataset at shutdown time, so I think it's simpler to make the class a singleton rather than make all its fields static.

I prefer using singleton objects with non-static methods in general (vs static methods with static members), since you can still have "real" static methods that don't rely on any class state.

integration/src/main/java/scripts/api/TdrDatasetsApi.java
pshapiro4broad: I tried using uuid at first but hyphens are disallowed so I switched to millis. Do you really think we'll have a collision of milliseconds in practice? That seems unlikely, but I could remove hyphens from a UUID instead of using milliseconds.

integration/src/main/java/scripts/api/TdrDatasetsApi.java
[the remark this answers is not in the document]
tcjiang9: Sounds good. Although I think it's worth leaving a comment explaining why we have a singleton, as without knowledge of the TDR locking issue, it's a puzzling choice.

integration/src/main/java/scripts/api/SnapshotsApi.java
pshapiro4broad: The same reason as not extending `DatasetsApi`; since in general I think it's better to encapsulate and so the caller doesn't try to call `deleteSnapshot()` directly. If it turned out that there were many APIs that needed forwarding I may have left it as-is to avoid all the forwarding APIs. But in this case there were only a few APIs that were forwarded.

petesantos (approved):

tcjiang9 (approved):

PR #91, opened 2022-05-25
DC-345: Workspace operations integration test

pshapiro4broad:
Work in progress.

This enables the test, and works, but the hardcoded billing account needs to be replaced with dynamic account creation/deletion so the test runs in all environments.

### DIFF
--- a/integration/src/main/java/scripts/client/RawlsClient.java
+++ b/integration/src/main/java/scripts/client/RawlsClient.java
@@ -6,8 +6,22 @@ import bio.terra.testrunner.runner.config.ServerSpecification;
 import bio.terra.testrunner.runner.config.TestUserSpecification;
 import com.google.auth.oauth2.GoogleCredentials;
 import java.io.IOException;
+import java.util.Objects;
 
 public class RawlsClient extends ApiClient {
+
+  boolean deleteWorkspaceWorkaround;
+
+  @Override
+  public String selectHeaderAccept(String[] accepts) {
+    // This workaround is necessary because the rawls openAPI spec doesn't match the endpoint's
+    // behavior. Without this change, calling deleteWorkspace() will fail with a 406 status.
+    if (deleteWorkspaceWorkaround) {
+      return "text/plain";
+    }
+    return super.selectHeaderAccept(accepts);
+  }
+
   /**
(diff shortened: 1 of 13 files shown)


integration/src/main/java/scripts/client/RawlsClient.java
```diff
@@ -1,13 +1,78 @@
...
+import java.util.Objects;
+import java.util.UUID;
+import java.util.stream.Stream;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class RawlsClient {
+  private static final Logger log = LoggerFactory.getLogger(RawlsClient.class);
+
+  private static final String BILLING_ACCOUNT = "billingAccounts/00708C-45D19D-27AAFA";
```
fboulnois, line 26: Should we make this configurable with an `integration.billintAccount` property or similar?

fboulnois, line 26: Should we make this configurable with an `integration.billingAccount` property or similar?

fboulnois (commented):
Overall one minor suggestion:

fboulnois (approved):

integration/src/main/java/scripts/client/RawlsClient.java
pshapiro4broad: There is a built-in property for this that I considered using, `TestConfiguration.billingAccount`. But if I used that I'd have to add the same billing account to all our test configurations, which seems wasteful as we only have one billing account for all test configurations (and all test environments). If we had different billing accounts then I think it would make sense to do this.

If you look for usages of this billing account in our source code, the main user is TDR, via this test config property. https://github.com/search?p=1&q=org%3ADataBiosphere+00708C-45D19D-27AAFA&type=Code

Within the broadinstitute org, the main usage is in a general repo that's used to configure tests in different repos. This is an older setup so we won't be using it for catalog. See https://github.com/broadinstitute/firecloud-automated-testing/search?q=00708C-45D19D-27AAFA

pshapiro4broad: That is to say, I'm not against putting it in a config file, but I don't know where it would make sense to put it, since it is used for all tests in all environments.

tcjiang9 (approved):

PR #93, opened 2022-06-06
DC-344: Add script to manage test workspaces

fboulnois:
Also remove the old script to run manual tests


scripts/catalog-tests/manage-test-workspaces.sh
```diff
@@ -0,0 +1,61 @@
...
+
+delete_action() {
+  delete_workspace
+  delete_billing_project
+}
+
+AUTH_TOKEN=$(printf "Authorization: Bearer %s" "$(gcloud auth print-access-token)")
+BILLING_ACCT="billingAccounts/00708C-45D19D-27AAFA"
+PROJECT_NAME="catalog_test_project"
+WORKSPACE_NAME="catalog_test_workspace_discoverable"
```
fboulnois, line 47: I will use a for loop to create these instead.

scripts/catalog-tests/manage-test-workspaces.sh
```diff
@@ -0,0 +1,61 @@
...
+  curl -X 'DELETE' -H "${AUTH_TOKEN}" \
+    "${RAWLS_ENV}/api/workspaces/${PROJECT_NAME}/${WORKSPACE_NAME}"
+}
+
+delete_action() {
+  delete_workspace
+  delete_billing_project
+}
+
+AUTH_TOKEN=$(printf "Authorization: Bearer %s" "$(gcloud auth print-access-token)")
```
pshapiro4broad, line 44: Is there a reason to use `printf` here? You can expand the subshell output directly into the string:
```suggestion
AUTH_TOKEN="Authorization: Bearer $(gcloud auth print-access-token)"
```

scripts/catalog-tests/manage-test-workspaces.sh
```diff
@@ -0,0 +1,61 @@
...
+WORKSPACE_NAME="catalog_test_workspace_discoverable"
+
+if [ $# -lt 2 ]; then
+  usage
+fi
+case $1 in
+  dev | perf | alpha | staging) ;;
+  *) usage ;;
+esac
+RAWLS_ENV=$(printf "https://rawls.dsde-%s.broadinstitute.org/api" "$1")
```
pshapiro4broad, line 56: This can also be done without `printf`
```suggestion
RAWLS_ENV="https://rawls.dsde-${1}.broadinstitute.org/api"
```

scripts/catalog-tests/manage-test-workspaces.sh
```diff
@@ -0,0 +1,61 @@
...
+AUTH_TOKEN=$(printf "Authorization: Bearer %s" "$(gcloud auth print-access-token)")
+BILLING_ACCT="billingAccounts/00708C-45D19D-27AAFA"
+PROJECT_NAME="catalog_test_project"
+WORKSPACE_NAME="catalog_test_workspace_discoverable"
+
+if [ $# -lt 2 ]; then
+  usage
+fi
+case $1 in
+  dev | perf | alpha | staging) ;;
```
pshapiro4broad, line 53: It doesn't hurt to do this but I think currently we only need workspaces in dev.

pshapiro4broad (approved):
A few comments but overall this looks good.

One thing I was thinking of is that `gcloud auth print-access-token` returns the currently select account's auth token. This may not be what a user expects; I've also used `gcloud auth application-default print-access-token` but I'm not sure if that's any better. Maybe the script could print the email for the account it's using? I think this would work:
```
EMAIL=$(gcloud auth list --format=json | jq -r '.[] | select(.status=="ACTIVE") | .account')
echo "Running commands as ${EMAIL}"
```

scripts/catalog-tests/manage-test-workspaces.sh
```diff
@@ -0,0 +1,61 @@
...
+}
+
+delete_action() {
+  delete_workspace
+  delete_billing_project
+}
+
+AUTH_TOKEN=$(printf "Authorization: Bearer %s" "$(gcloud auth print-access-token)")
+BILLING_ACCT="billingAccounts/00708C-45D19D-27AAFA"
+PROJECT_NAME="catalog_test_project"
```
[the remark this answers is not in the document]
fboulnois: Yes, I think we only want to use one project for test workspaces.

scripts/catalog-tests/manage-test-workspaces.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/sh
+
+usage() {
+  echo "usage: $0 <env> <create/delete>"
+  exit 1
+}
+
+create_billing_project() {
+  curl -X 'POST' \
```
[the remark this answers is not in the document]
fboulnois: I had this initially, but it might be convenient to try to continue even if the project step fails due to it already existing. But perhaps not.

scripts/catalog-tests/manage-test-workspaces.sh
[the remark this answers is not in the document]
fboulnois: I also revised the script to be able to delete workspaces and the workspace project separately.

philshap (approved):
Looks good!

pshapiro4broad (approved):
Whoops. I mean *clears throat* Looks good!

petesantos (approved):

PR #94, opened 2022-06-10
DC-395: update spotless plugin and remove workaround in gradle.properties

pshapiro4broad:
Also updated many other plugin versions to the latest version.

### DIFF
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -10,16 +10,16 @@ repositories {
 }
 
 dependencies {
-    implementation 'com.diffplug.spotless:spotless-plugin-gradle:6.5.0'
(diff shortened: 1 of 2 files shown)


fboulnois (approved):
Glad this was finally fixed!

s-rubenstein (approved):
LGTM. Out of curiosity, how did you get the specific version numbers? Are they just latest?

pshapiro4broad:
> Out of curiosity, how did you get the specific version numbers? Are they just latest?

Yes, I updated to latest for all except spotbugs, which had a major rev and I didn't want to take that on, in case it caused problems. Our spotbugs config is not great anyway, see DC-322.

PR #96, opened 2022-06-15
[DC-377] Remove unused imports and add last modified

s-rubenstein:


fboulnois (approved):
Good idea.

scripts/workspace-conversion/workspace-to-dataset.py
```diff
@@ -302,6 +302,7 @@ def generate_catalog_metadata(workspace):
         "dct:title": wsAttributes.pop("library:datasetName", None),
         "dct:version": wsAttributes.pop("library:datasetVersion", None),
         "dct:description": wsAttributes.pop("library:datasetDescription", None),
+        "dct:modified": datetime.now().strftime("%Y-%m-%dT%H:%M:%S"),
```
pshapiro4broad, line 305: Is there a reason to not use `workspace["workspace"]["lastModified"]` here? That seems like it contains the last modified date for the workspace.

pshapiro4broad (approved):
Looks OK to me. I notice we're not following PEP8 for imports https://peps.python.org/pep-0008/#imports

scripts/workspace-conversion/workspace-to-dataset.py
s-rubenstein: I thought this would be the time the catalog entry was last modified, not the time the workspace/underlying data was last modified.

s-rubenstein: Talked offline, moved.

PR #98, opened 2022-06-17
[DC-403] Add script to publish datasets

s-rubenstein:


scripts/publish-metadata.py
```diff
@@ -0,0 +1,107 @@
...
+    create_dataset(args.environment, access_token, args.storage_system, args.storage_system_id, data)
+
+    auth_as_user(current_user)
+
+
+
+if __name__ == "__main__":
+    current_user = subprocess.run(['gcloud', 'config', 'get-value', 'account'], check=True, capture_output=True).stdout.strip().decode("utf-8")
+    try:
+        parser = argparse.ArgumentParser(
```
s-rubenstein, line 66: I wanted to use the argparser because I was noticing the environment variable approach we have taken in other scripts has bled into my environment. I don't really like that sort of side effect, and I know how to use this.

scripts/publish-metadata.py
```diff
@@ -0,0 +1,107 @@
...
+def main(args, current_user):
+    print(args)
+    access_token = generate_access_token(args.user)
+    print(access_token)
+    # Opening JSON file
+    with open(args.metadata_file, 'r') as f:
+        data = json.load(f)
+    create_dataset(args.environment, access_token, args.storage_system, args.storage_system_id, data)
+
+    auth_as_user(current_user)
```
s-rubenstein, line 59: Another side effect I noticed was that we left you logged in as the account specified, which once again kinda causes changes to your system to be left behind. I thought it might be good to auth back when done. (Especially for folks who use their firecloud.org account in prod to not accidentally stay authed)

scripts/publish-metadata.py
```diff
@@ -0,0 +1,120 @@
...
+    try:
+        parser = argparse.ArgumentParser(
+            description=DESCRIPTION, formatter_class=argparse.RawTextHelpFormatter
+        )
+        parser.add_argument(
+            "--environment",
+            type=str,
+            choices=["prod", "dev", "alpha", "staging", "perf"],
+            default="dev",
+            help="firecloud environment to create dataset in",
```
pshapiro4broad, line 92: Maybe this is the `terra` or `catalog` environment?

scripts/publish-metadata.py
```diff
@@ -0,0 +1,120 @@
...
+        "Content-Type": "application/json",
+    }
+
+    request = {
+        "storageSystem": str(storage_system),
+        "storageSourceId": str(storage_system_id),
+        "catalogEntry": json.dumps(entry),
+    }
+    print(headers)
+    response = requests.post(
```
pshapiro4broad, line 46: If the dataset already exists, `post` will fail and you need to use `put` instead. Or use `get` before `post` to see if it already exists. Maybe something like https://github.com/DataBiosphere/terra-data-catalog/blob/main/scripts/catalog-tests/generate-tdr-snapshot-metadata.py#L68

Not sure if it's time to think about avoiding code duplication between this script and `generate-tdr-snapshot-metadata.py` and `ingest-hca-collection.py`. At some point we'll probably want to refactor these scripts.

pshapiro4broad: And/or we could look at using the python client codegen from swagger for this code.

pshapiro4broad (approved):
Looks OK to me, although I wonder if there's a way to use this from our other scripts as well.

scripts/publish-metadata.py
s-rubenstein: Oh, yup!

scripts/publish-metadata.py
s-rubenstein: TOL: 
So, this is specifically intended to just handle `post`. I didn't want to handle updating too, because I didn't know if that did smart updates. However, it looks like its always a full replacement? So I can change this to be `publish-or-update` metadata.

My other concern is that in order to `get`, we would need the data-catalog id, which we don't necessarily have at this point. So we would need to `list`, `filter` by `storageSourceId`, and then `put`. This is doable. It is a bit more work, but that seems worth it to me.

pshapiro4broad: > However, it looks like its always a full replacement?

Yes, that's right.

> My other concern is that in order to get, we would need the data-catalog id, which we don't necessarily have at this point. So we would need to list, filter by storageSourceId, and then put. This is doable. It is a bit more work, but that seems worth it to me.

Yeah my text didn't match the code I linked to, which does the `list` and filters the result.

s-rubenstein: We can't do that. We don't return `storageSourceId` with our `list` endpoint. So we would need to add that information (as well as storage system) to our `list` response before we could detect if it exists already/figure out which dataset is the one we care about.

s-rubenstein: Hmm… looking at the code you linked actually, looks like we do have `dct:identifier` for TDR entries, but not for workspace entries yet. I'll add that to the list. However, either way, we are missing the storage system in both cases, so we'd need to add that to our response.

s-rubenstein: We can't retry after error on POST, because we have no way to tell what to PUT. You need to get the data catalog ID and there's no way to derive that information from the storage source id and storage system. We could potentially add that information, but that would be a service change to add more information to our return objects.

pshapiro4broad: It looks like HTTP status 409 (conflict) can be used here. If our POST API receives data which would result in a duplicate entry, we can return a 409 status response with the URL to the existing resource (possibly in the `Content-Location:` header). See https://www.rfc-editor.org/rfc/rfc9110.html#name-409-conflict.

s-rubenstein: I hesitate to block this on updating our endpoint to return a better error message though. I can file a ticket to add the url to the 409 error that you get back when a dataset for a particular storage id has already been created.

pshapiro4broad: I agree, let's not block this PR on this feature. We can go back and add support once we add the support in the service code. In the meantime we're not yet running this in bulk so I think it's OK to require manual intervention in the case of duplicate entries.

scripts/publish-metadata.py
```diff
@@ -0,0 +1,117 @@
+import argparse
+import sys
+import requests
+import json
+import subprocess
+
+DESCRIPTION = """
```
fboulnois, line 7: I would also explain that the `requests` library is necessary, and that `gcloud` needs to be installed.

scripts/publish-metadata.py
```diff
@@ -0,0 +1,117 @@
...
+        "Authorization": f"Bearer {access_token}",
+        "Content-Type": "application/json",
+    }
+
+    request = {
+        "storageSystem": str(storage_system),
+        "storageSourceId": str(storage_source_id),
+        "catalogEntry": json.dumps(entry),
+    }
+    response = requests.post(
```
fboulnois, line 45: I highly recommend any `requests` installation occurs in a `virtualenv`, it avoids many of the issues with system packages. See the other scripts for how to do this.

fboulnois (commented):
A few suggestions:

scripts/publish-metadata.py
s-rubenstein: I can add a comment similar to what Christine did. I think that helpfully explains how to set up a venv while allowing users to use their own mechanisms to use `requests` if they would like to do otherwise.

scripts/publish-metadata.py
```diff
@@ -0,0 +1,130 @@
...
+import subprocess
+
+DESCRIPTION = """
+Uploads a given block of metadata to the Terra Data Catalog
+
+You must have the following installed to run this script:
+# The Python library `requests`
+# The Google Cloud SDK
+
+# To set up a venv with requests installed, follow the below instructions
```
fboulnois, line 14:
```suggestion
# To set up a venv with requests installed, follow the instructions below:
```

fboulnois (approved):

PR #99, opened 2022-06-17
[DC-377] Add workspace id as dct:identifier

s-rubenstein:


fboulnois (approved):
👍

pshapiro4broad (commented):
The more I think about this, the more I think the TDR code is wrong. My guess is the TDR ingest code was directly ported from the original TDR scripts, where `dct:identifier` was the same as the TDR snapshot ID. In catalog I don't think this is correct anymore. I think a catalog entry should use the catalog ID for its `dct:identifier`, since the identifier is supposed to identify the item in the data catalog, not in its underlying storage system (which may not exist, in the case of an external dataset).

Assuming that `dct:identifier` is the catalog's ID, we then have to think about use cases where a user might want the storage system's ID. Before we do something here, I think we want to figure out what the use cases are, and if there's other ways to address them.

pshapiro4broad:
> So, if we make dct:identifier be the catalog ID, then it becomes a duplicate of id. I am inclined to just get rid of the field (or the id field) if we are going to do that, and not duplicate values.

Yes, I think we should remove `id` and only send `dct:identifier`.

> I think there could be use cases where a user might want to go to the underlying data source, but we want to discourage that in general. Ideally users don't need to know how to query the Data Repo. The one situation I could see where a user would want to do that, and we would want to encourage it, is with power users. But I don't know whether power users would just go directly to the storage source without going through the catalog or not.

There is `dcat:accessURL`. Although maybe this should also point at the catalog, not back to TDR or terra-ui. We were thinking that we want a way to store and show the workspace link for a workspace dataset, so we will need some place to store the storage system URL, if it's not `cat:accessURL`.

s-rubenstein:
I'm going to push back on making that change right now. Currently, we have a number of snapshots (albeit in just dev) that do it the other way. I could see us making this change, but we would need to figure out that backfill. We could, potentially, just make the change and have it override the metadata, but I am a bit concerned with overwriting user defined metadata before we have a schema for them.

We could just not do this here, and then when we have our schema, make it clear that the `dct:identifier` is a reserved field. But as long as users can put whatever they want in `dct:identifier`, it makes me a little uncomfortable to blow away that data that they explicitly included.

s-rubenstein:
That said, lets skip this for now, ignore `dct:identifier`, and figure this out with the schema discussion later.

PR #103, opened 2022-06-28
DC-422: Pin black version for build

pshapiro4broad:
Letting the black version float can cause a build failure when the default version installed by `pip install black` doesn't match the version specified by the spotless python plugin.


fboulnois (approved):

s-rubenstein (approved):
LGTM

PR #110, opened 2022-07-14
[DC-419] Admins can list all datasets, even if they don't have permissions on originating resource

s-rubenstein:
Tested to make sure things weren't double listed.

### DIFF
--- a/service/src/main/java/bio/terra/catalog/service/DatasetService.java
+++ b/service/src/main/java/bio/terra/catalog/service/DatasetService.java
@@ -75,15 +75,15 @@ public class DatasetService {
       node.set("id", TextNode.valueOf(dataset.id().toValue()));
       return node;
     }
-  }
 
-  private ObjectNode toJsonNode(String json) {
-    try {
-      return objectMapper.readValue(json, ObjectNode.class);
-    } catch (JsonProcessingException e) {
-      // This shouldn't occur, as the data stored in postgres must be valid JSON, because it's
-      // stored as JSONB.
-      throw new IllegalMetadataException(e);
+    private ObjectNode toJsonNode(String json) {
+      try {
+        return objectMapper.readValue(json, ObjectNode.class);
+      } catch (JsonProcessingException e) {
+        // This shouldn't occur, as the data stored in postgres must be valid JSON, because it's
+        // stored as JSONB.
+        throw new IllegalMetadataException(e);
+      }
     }
   }
 
--- a/service/src/main/java/bio/terra/catalog/service/DatasetService.java
+++ b/service/src/main/java/bio/terra/catalog/service/DatasetService.java
@@ -24,6 +24,7 @@ import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.fasterxml.jackson.databind.node.TextNode;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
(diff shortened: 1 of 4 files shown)


service/src/main/java/bio/terra/catalog/service/dataset/DatasetDao.java
```diff
@@ -132,4 +132,9 @@ public List<Dataset> find(StorageSystem storageSystem, Collection<String> ids) {
     var params = Map.of("storageSystem", String.valueOf(storageSystem), "ids", ids);
     return jdbcTemplate.query(sql, params, new DatasetMapper());
   }
+
+  @ReadTransaction
+  public List<Dataset> listAllDatasets() {
+    return jdbcTemplate.query("SELECT * FROM dataset", new DatasetMapper());
```
fboulnois, line 138:
```suggestion
    String sql = "SELECT * FROM dataset";
    return jdbcTemplate.query(sql, new DatasetMapper());
```

fboulnois (approved):
lgtm, I guess we probably want some unit tests including `DatasetWithAccessLevel`?

service/src/main/java/bio/terra/catalog/service/dataset/DatasetDao.java
s-rubenstein: Because it doesn't have params that need mapping, I didn't see a need to abstract it out to an intermediate variable. I don't object to it, it just seems unnecessary in this case. (All the other methods do have variables passed in)

s-rubenstein:
For `datasetWithAccessLevel`, codecov actually thinks it is handled because it is exercised as a part of the listDatasets endpoint. I think that it is actually mostly implementation detail of our methods? But I could see wanting more tests if we start treating it more as a first order class, as opposed to a helper class. What are your thoughts on the distinction there?

petesantos (approved):

service/src/main/java/bio/terra/catalog/service/DatasetService.java
```diff
@@ -55,49 +56,76 @@ public IllegalMetadataException(Throwable cause) {
...
   }
 
-  private ObjectNode toJsonNode(String json) {
-    try {
-      return objectMapper.readValue(json, ObjectNode.class);
-    } catch (JsonProcessingException e) {
-      // This shouldn't occur, as the data stored in postgres must be valid JSON, because it's
-      // stored as JSONB.
-      throw new IllegalMetadataException(e);
+  private class DatasetWithAccessLevel {
```
pshapiro4broad, line 59: I think `record` would work well here

service/src/main/java/bio/terra/catalog/service/DatasetService.java
```diff
@@ -55,49 +56,76 @@ public IllegalMetadataException(Throwable cause) {
...
-    return convertSourceObjectsToDatasetResponses(roleMap, StorageSystem.TERRA_WORKSPACE);
+    return convertSourceObjectsToDatasetsWithAccessLevel(roleMap, StorageSystem.TERRA_WORKSPACE);
   }
 
   public DatasetsListResponse listDatasets(AuthenticatedUserRequest user) {
+    List<DatasetWithAccessLevel> datasets = new ArrayList<>();
+    datasets.addAll(collectWorkspaceDatasets(user));
+    datasets.addAll(collectDatarepoDatasets(user));
+    if (samService.hasGlobalAction(user, SamAction.READ_ANY_METADATA)) {
+      datasets.addAll(
```
pshapiro4broad, line 115: In the admin user case, we make three database calls but only one is needed. Even for the non-workspace case, we could be doing a single database call for all storage systems. When we add external dataset support we will need to revisit this anyway so there's no rush on changing it now.

service/src/main/java/bio/terra/catalog/service/DatasetService.java
```diff
@@ -55,49 +56,76 @@ public IllegalMetadataException(Throwable cause) {
...
+          datasetDao.listAllDatasets().stream()
+              .map(dataset -> new DatasetWithAccessLevel(dataset, DatasetAccessLevel.READER))
+              .filter(
+                  datasetWithAccessLevel ->
+                      !datasets.stream()
+                          .map(datasetInList -> datasetInList.getDataset().id())
+                          .toList()
+                          .contains(datasetWithAccessLevel.getDataset().id()))
+              .toList());
```
pshapiro4broad, line 124: I think `x.addAll(y.stream().toList())` is less memory efficient than `y.stream().forEach(x::add)`

PR #111, opened 2022-07-21
[DC-346] workspace permissions integration test

s-rubenstein:


integration/src/main/java/scripts/testscripts/WorkspacePermissionOperations.java
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  private void testNoPermissions() throws Exception {
+    clearTestWorkspacePermissions();
+    // verify the user cannot access dataset without any permissions
+    assertThrows(ApiException.class, () -> userDatasetsApi.getDataset(adminTestDatasetId));
+    assertThat(
+        userDatasetsApi.getApiClient().getStatusCode(), is(HttpStatusCodes.STATUS_CODE_NOT_FOUND));
+  }
+  // TODO (DC-446): Fix with https://broadworkbench.atlassian.net/browse/DC-446
```
pshapiro4broad, line 98: Since there's a JIRA for this, can this test code go into the JIRA as well? I think that would be better than leaving commented out code in the source.

integration/build.gradle
```diff
@@ -18,6 +18,9 @@ dependencies {
     implementation 'org.junit.jupiter:junit-jupiter-api'
     implementation 'org.hamcrest:hamcrest'
 
+    implementation 'org.glassfish.jersey.core:jersey-client'
```
pshapiro4broad, line 21: Is this dependency required? I would expect that `jersey-jdk-connector` would be sufficient by itself.

integration/src/main/java/scripts/testscripts/WorkspacePermissionOperations.java
```diff
@@ -0,0 +1,173 @@
...
+import com.google.api.client.http.HttpStatusCodes;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.UUID;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import scripts.client.CatalogClient;
+import scripts.client.RawlsClient;
+
+public class WorkspacePermissionOperations extends TestScript {
```
pshapiro4broad, line 26: Would it be hard to test both storage systems in a single permissions test, in the way that we do for `DatasetOperations`? If they're testing the same cases, the common code could be shared and it would be easier to see that we're testing the same things for both storage systems.

integration/src/main/java/scripts/testscripts/WorkspacePermissionOperations.java
```diff
@@ -0,0 +1,173 @@
...
+    adminRawlsClient.updateWorkspaceAcl(
+        List.of(new WorkspaceACLUpdate().email(regularUser.userEmail).accessLevel(policy)),
+        adminTestWorkspace);
+  }
+
+  private void clearTestWorkspacePermissions() throws Exception {
+    log.info(
+        "Clearing acl permissions for {}/{}",
+        adminTestWorkspace.getNamespace(),
+        adminTestWorkspace.getName());
```
pshapiro4broad, line 132: I don't know if we need to log this. If we do, maybe it should be done inside `RawlsClient.updateWorkspaceAcl()` so both setting and unsetting the ACL is logged.

integration/src/main/java/scripts/testscripts/WorkspacePermissionOperations.java
```diff
@@ -0,0 +1,173 @@
...
+    CreateDatasetRequest request = datasetRequestForWorkspace(userTestWorkspace.getWorkspaceId());
+    adminCreateDataset(request);
+  }
+
+  private void setTestWorkspacePermissionForRegularUser(String policy) throws Exception {
+    // first clear the shared snapshot of policy state caused by previous tests
+    adminRawlsClient.updateWorkspaceAcl(
+        List.of(new WorkspaceACLUpdate().email(regularUser.userEmail).accessLevel(policy)),
+        adminTestWorkspace);
+  }
```
pshapiro4broad, line 126: If `adminRawlsClient.updateWorkspaceAcl()` supported an API like `adminRawlsClient.updateWorkspaceAcl(adminTestWorkspace, regularUser.userEmail, WorkspaceAccessLevel.READER)` then maybe this method wouldn't be needed anymore, since most of the mess of making the ACL list would be hidden from this code.

integration/src/main/java/scripts/testscripts/WorkspacePermissionOperations.java
```diff
@@ -0,0 +1,173 @@
...
+  public void cleanup(List<TestUserSpecification> testUsers) throws Exception {
+    for (var datasetId : datasetIds) {
+      adminDatasetsApi.deleteDataset(datasetId);
+      log.info("deleted dataset {}", datasetId);
+    }
+    if (adminTestWorkspace != null) {
+      log.info(
+          "deleting workspace {}/{}",
+          adminTestWorkspace.getNamespace(),
+          adminTestWorkspace.getName());
```
pshapiro4broad, line 162: I think this information is already logged in `RawslClient.deleteWorkspace()`, and doesn't need to be logged here too.

pshapiro4broad (approved):
I have a few suggestions, although the code looks good overall.

integration/src/main/java/scripts/testscripts/WorkspacePermissionOperations.java
s-rubenstein: That sounds great! I dropped the code into a comment on the ticket.

integration/src/main/java/scripts/testscripts/WorkspacePermissionOperations.java
s-rubenstein: I like that quite a bit. I'll see if can move it over

integration/src/main/java/scripts/client/RawlsClient.java
```diff
@@ -39,6 +42,12 @@ private WorkspacesApi createWorkspacesApi(String basePath, TestUserSpecification
       throws IOException {
     var workspaceApiClient =
         new ApiClient() {
+          @Override
+          protected void performAdditionalClientConfiguration(ClientConfig clientConfig) {
+            super.performAdditionalClientConfiguration(clientConfig);
+            clientConfig.connectorProvider(new JdkConnectorProvider());
+          }
+
```
fboulnois, line 50: Did this change end up fixing the issue we were facing?

fboulnois (approved):
I like @pshapiro4broad 's suggestions, otherwise lgtm

integration/src/main/java/scripts/client/RawlsClient.java
s-rubenstein: Yes it did!

integration/src/main/java/scripts/testscripts/DatasetPermissionOperations.java
```diff
@@ -67,37 +77,82 @@ public void setup(List<TestUserSpecification> testUsers) throws Exception {
...
+    userDatasetsApi.getDataset(adminTestSnapshotDatasetId);
     assertThat(userDatasetsApi.getApiClient().getStatusCode(), is(HttpStatusCodes.STATUS_CODE_OK));
   }
 
-  private void testDiscovererPermissions() throws Exception {
+  private void testReaderPermissionsForWorkspaceDataset() throws Exception {
+    setTestWorkspacePermissionForRegularUser(WorkspaceAccessLevel.READER);
+    // User cannot create a catalog entry on snapshot when user only has reader access on snapshot
+    CreateDatasetRequest request =
+        datasetRequest(adminTestWorkspace.getWorkspaceId(), StorageSystem.WKS);
```
pshapiro4broad, line 123: This is better, but would it be possible to make a version that could be used for both storage systems? What if this method accepted a `datasetRequest` argument and a `adminTestDatasetId` argument?

PR #112, opened 2022-07-25
DC-450: Remove jstl library

fboulnois:
This library doesn't seem to be necessary to build or run the project and has a vulnerability.


pshapiro4broad (approved):
I wonder how the templating works without this. I thought that maybe it was still bringing in `jstl` via a transitive dependency, but that doesn't seem to be the case. Either way it's nice to remove a dependency.

s-rubenstein (approved):
LGTM

PR #113, opened 2022-07-26
DC-456: Add export endpoint to openapi.yml

fboulnois:


common/openapi.yml
```diff
@@ -145,6 +145,25 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorReport'
+  /api/v1/datasets/{id}/export:
+    post:
```
fboulnois, line 149: Is a post the right call here?

common/openapi.yml
```diff
@@ -145,6 +145,25 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorReport'
+  /api/v1/datasets/{id}/export:
```
fboulnois, line 148: Thoughts about this endpoint?

common/openapi.yml
```diff
@@ -145,6 +145,25 @@ paths:
...
+        - $ref: '#/components/parameters/Id'
+      operationId: exportDataset
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/CreateExportRequest'
+        required: true
+      responses:
+        '204':
```
fboulnois, line 162: Do we need to capture a response on success?

common/openapi.yml
```diff
@@ -145,6 +145,25 @@ paths:
...
+          description: The data was successfully exported into a workspace
+          content: { }
+        '404':
```
fboulnois, line 165: Do we need to capture a more specific error?

common/openapi.yml
```diff
@@ -358,6 +377,24 @@ components:
...
       enum: ['tdr', 'wks', 'ext']
 
+    WorkspaceId:
+      description: The workspace UUID
+      type: object
+      required: [ id ]
+      properties:
+        id:
+          type: string
+          format: uuid
```
fboulnois, line 387: I was thinking of strongly typing the WorkspaceId, so it couldn't accidentally be defined as another type of uuid.

common/openapi.yml
pshapiro4broad: I think `export` makes sense. (vs `link` which we aren't planning to do at this point).

common/openapi.yml
pshapiro4broad: I think so, I think it makes more sense than `PUT` for this operation.

common/openapi.yml
pshapiro4broad: Probably not.. if we were creating a workspace then returning the ID would make sense, but not if we're accepting the ID.

common/openapi.yml
```diff
@@ -145,6 +145,25 @@ paths:
...
               schema:
                 $ref: '#/components/schemas/ErrorReport'
+  /api/v1/datasets/{id}/export:
+    post:
+      summary: Export the dataset data into a workspace
+      tags: [ datasets ]
+      parameters:
+        - $ref: '#/components/parameters/Id'
+      operationId: exportDataset
+      requestBody:
```
pshapiro4broad, line 155: For a required argument, is there a reason to use a body instead of a path argument? It seems equivalent to me, unless the body is large and requires a multipart attachment. Using a path argument makes it a little easier to use from scripts and swagger, IMO.

common/openapi.yml
s-rubenstein: I think we probably want to throw a 403 if a user has 'Discoverer' but not 'Reader' access?

s-rubenstein (approved):
LGTM (Pending change)

tcjiang9 (approved):

common/openapi.yml
pshapiro4broad: Maybe, but TDC shouldn't do any permission check here, it should leave it up to the storage system to do.

s-rubenstein: Agreed that we shouldn't call to any permission check endpoint, but re: what the users of the API should expect, we want to provide possible error cases in our documentation.

PR #114, opened 2022-07-28
DC-457, DC-459: Add export to workspace

fboulnois:


service/src/main/java/bio/terra/catalog/rawls/RawlsService.java
```diff
@@ -99,4 +103,45 @@ public SystemStatusSystems status() {
...
+      // possible bug: empty entityType and entityNames copies all entities
+      EntityCopyDefinition body =
+          new EntityCopyDefinition()
+              .sourceWorkspace(workspaceNameSource)
+              .destinationWorkspace(workspaceNameDest)
+              .entityType("")
+              .entityNames(List.of());
+      entitiesApi(user).copyEntities(body, false);
+    } catch (ApiException e) {
+      // todo
```
fboulnois, line 144: What should we do if we get an error? There are many errors that are possible, but one obvious one is what happens if you export to workspace with entities that already exist in the destination workspace?

s-rubenstein (commented):
Do we have a ticket for integration tests? That should probably be done in a follow up PR.

service/src/main/java/bio/terra/catalog/service/DatasetService.java
```diff
@@ -252,4 +252,17 @@ public DatasetPreviewTable getDatasetPreview(
     var dataset = datasetDao.retrieve(datasetId);
     return generateDatasetTablePreview(user, dataset, tableName);
   }
+
+  public void exportDataset(
+      AuthenticatedUserRequest user, DatasetId datasetId, String workspaceId) {
+    var dataset = datasetDao.retrieve(datasetId);
+    ensureActionPermission(user, dataset, SamAction.READ_ANY_METADATA);
```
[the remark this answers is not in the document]
fboulnois: I'm not totally sure about this one, but I implemented the others.

service/src/main/java/bio/terra/catalog/api/DatasetApiController.java
```diff
@@ -88,4 +88,10 @@ public ResponseEntity<DatasetPreviewTable> getDatasetPreviewTable(UUID id, Strin
         .cacheControl(CacheControl.noStore())
         .body(datasetService.getDatasetPreview(getUser(), new DatasetId(id), tableName));
   }
+
+  @Override
+  public ResponseEntity<Void> exportDataset(UUID datasetId, UUID workspaceId) {
+    datasetService.exportDataset(getUser(), new DatasetId(datasetId), workspaceId.toString());
```
pshapiro4broad, line 94: Is there a reason to convert the workspace ID into a string at this layer? Would it be incorrect to make the `DatasetService` API use UUIDs? I think it would be nice to use a more specific type in our code if possible.

service/src/main/java/bio/terra/catalog/rawls/RawlsService.java
```diff
@@ -99,4 +103,44 @@ public SystemStatusSystems status() {
...
+  public void exportWorkspaceDataset(
+      AuthenticatedUserRequest user, String workspaceIdSource, String workspaceIdDest) {
+    try {
+      // build source name
+      WorkspaceDetails workspaceDetailsSource =
+          workspacesApi(user).getWorkspaceById(workspaceIdSource, List.of()).getWorkspace();
+      WorkspaceName workspaceNameSource =
+          new WorkspaceName()
+              .namespace(workspaceDetailsSource.getNamespace())
+              .name(workspaceDetailsSource.getName());
```
pshapiro4broad, line 120: Would it be worth making a static method that converted from `WorkspaceDetails` to `WorkspaceName` to avoid the code duplication?

service/src/main/java/bio/terra/catalog/rawls/RawlsService.java
```diff
@@ -99,4 +103,44 @@ public SystemStatusSystems status() {
...
+              .name(workspaceDetailsDest.getName());
+
+      // possible bug: empty entityType and entityNames copies all entities
+      EntityCopyDefinition body =
+          new EntityCopyDefinition()
+              .sourceWorkspace(workspaceNameSource)
+              .destinationWorkspace(workspaceNameDest)
+              .entityType("")
+              .entityNames(List.of());
+      entitiesApi(user).copyEntities(body, false);
```
pshapiro4broad, line 137: This ignores `EntityCopyResponse` which may contain conflicts information, I assume generated when exporting into a workspace that already contains some entities. I don't think we should hold up this work to handle conflicts but maybe we should have a ticket to look at adding this support in the future?

service/src/main/java/bio/terra/catalog/service/DatasetService.java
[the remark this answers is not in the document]
pshapiro4broad: I agree that we shouldn't do a permission check here. At best it's duplication of effort, where we're doing the same check as the service; at worst we might prevent an operation that the service would allow.

service/src/test/java/bio/terra/catalog/api/DatasetApiControllerTest.java
```diff
@@ -188,4 +191,14 @@ void getDatasetPreviewTable() throws Exception {
         .andExpect(content().contentType(MediaType.APPLICATION_JSON))
         .andExpect(jsonPath("$.columns[0].name").value(columnName));
   }
+
+  @Test
+  void exportMetadata() throws Exception {
+    var datasetId = new DatasetId(UUID.randomUUID());
+    var workspaceId = UUID.randomUUID();
+    doNothing().when(datasetService).exportDataset(user, datasetId, String.valueOf(workspaceId));
```
pshapiro4broad, line 199: I assume this has the same effect as doing a `verify()` after the `perform()`. Is there a reason to prefer one form over the other? It looks like other tests in this class use `verify()` to do this.

service/src/test/java/bio/terra/catalog/datarepo/DatarepoServiceTest.java
```diff
@@ -165,4 +166,13 @@ void getPreviewTableDatarepoException() throws Exception {
     assertThat(t.getStatusCode(), is(HttpStatus.NOT_FOUND));
     assertThat(t.getMessage(), is("bio.terra.datarepo.client.ApiException: " + errorMessage));
   }
+
+  @Test
+  void getExportDatarepoException() {
```
pshapiro4broad, line 171: This is fine although it's only needed because this PR adds code to `DatasetService`. Alternatively you could make TDR snapshots the same as external (no-op in the switch) for now.

service/src/main/java/bio/terra/catalog/service/DatasetService.java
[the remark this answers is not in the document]
s-rubenstein: I think the general rule we've been following is to not check permissions if the underlying storage system is going to do it for us - it duplicates the check -> https://broadworkbench.atlassian.net/browse/DC-315 is the ticket that describes the why/history of that decision.

service/src/main/java/bio/terra/catalog/api/DatasetApiController.java
```diff
@@ -88,4 +88,10 @@ public ResponseEntity<DatasetPreviewTable> getDatasetPreviewTable(UUID id, Strin
         .cacheControl(CacheControl.noStore())
         .body(datasetService.getDatasetPreview(getUser(), new DatasetId(id), tableName));
   }
+
+  @Override
+  public ResponseEntity<Void> exportDataset(UUID datasetId, UUID workspaceId) {
```
s-rubenstein, line 93: Do you think it would be worth filing a tech debt ticket to change the other param names to be `datasetId` as well?

service/src/main/java/bio/terra/catalog/datarepo/DatarepoService.java
```diff
@@ -147,4 +148,9 @@ public SystemStatusSystems status() {
     }
     return result;
   }
+
+  public void exportDatarepoDataset(
```
pshapiro4broad, line 152: I think `exportSnapshot()` is a better name for this method as the operation isn't catalog dataset specific within the service code layer.

service/src/test/java/bio/terra/catalog/service/DatasetServiceTest.java
```diff
@@ -59,6 +65,8 @@ class DatasetServiceTest {
   private final AuthenticatedUserRequest user = mock(AuthenticatedUserRequest.class);
   private final DatasetId datasetId = new DatasetId(UUID.randomUUID());
   private final String sourceId = "sourceId";
+
+  private final String externalSourceId = "phs000000.v11.p8";
```
pshapiro4broad, line 69: I think in practice external source IDs will be UUIDs minted either within catalog or by catalog's callers. But for this test it shouldn't matter what the source ID is. Is there a reason why `DatasetServiceTest.sourceId` wouldn't work?

service/src/main/java/bio/terra/catalog/api/DatasetApiController.java
fboulnois: Done.

service/src/test/java/bio/terra/catalog/service/DatasetServiceTest.java
```diff
@@ -71,6 +79,15 @@ class DatasetServiceTest {
           StorageSystem.TERRA_DATA_REPO,
           metadata,
           null);
+
+  private final Dataset externalDataset =
```
pshapiro4broad, line 83: Not sure why these fields are static.. do they rely on data which changes in each test run? (looking at the change history, it's probably my fault that they weren't static before)

service/src/test/java/bio/terra/catalog/service/DatasetServiceTest.java
```diff
@@ -278,4 +295,42 @@ void getDatasetPreviewTable() {
         datasetPreviewTable.getColumns().get(0),
         is(new bio.terra.catalog.model.ColumnModel().name("column a").arrayOf(false)));
   }
+
+  @Test
+  void testExportDatarepoDataset() {
+    reset(datasetDao);
```
pshapiro4broad, line 301: I don't think this is necessary, as I'd expect `@Mock` objects to be re-initialized on every test run. Does it work without `reset()`? Looking at the code, this is also done in `testGetMetadataUsingTdrPermission()` but I think that use is also unnecessary and should've been removed when `mockDataset()` was added.

service/src/test/java/bio/terra/catalog/service/DatasetServiceTest.java
```diff
@@ -278,4 +295,42 @@ void getDatasetPreviewTable() {
...
+    doThrow(new BadRequestException("error"))
+        .when(datarepoService)
+        .exportDatarepoDataset(user, sourceId, workspaceId);
+    assertThrows(
+        BadRequestException.class,
+        () -> datasetService.exportDataset(user, datasetId, workspaceId));
+  }
+
+  @Test
+  void testExportExternalDataset() {
```
pshapiro4broad, line 314: This test doesn't appear to be specific to external datasets. It's verifying that `exportDataset()` is doing a permission check, which fails because the current user isn't an admin user. If you mock that in sam, then you can write a test that verifies that trying to export an external dataset is a no-op.
```
    when(samService.hasGlobalAction(user, SamAction.READ_ANY_METADATA)).thenReturn(true);
```

But I wouldn't recommend doing that as we haven't worked on external datasets yet so there's not much value to us to verify our APIs behavior when using external datasets.

So you could keep this test but make it use either the TDR or workspace dataset.

service/src/test/java/bio/terra/catalog/service/DatasetServiceTest.java
```diff
@@ -278,4 +295,42 @@ void getDatasetPreviewTable() {
...
+    reset(datasetDao);
+    when(datasetDao.retrieve(datasetId)).thenReturn(workspaceDataset);
+    when(rawlsService.getRole(user, workspaceSourceId)).thenReturn(DatasetAccessLevel.READER);
+    String workspaceId = String.valueOf(UUID.randomUUID());
+    datasetService.exportDataset(user, datasetId, workspaceId);
+
+    DatasetService mockDatasetService = mock(DatasetService.class);
+    doNothing().when(mockDatasetService).exportDataset(user, datasetId, workspaceId);
+    mockDatasetService.exportDataset(user, datasetId, workspaceId);
+    verify(mockDatasetService).exportDataset(user, datasetId, workspaceId);
```
pshapiro4broad, line 334: I don't think this is necessary with strict subbing. If it's not called, you should get a test failure because of the mock on line 332.

pshapiro4broad (approved):
Many comments, looks good overall.

service/src/main/java/bio/terra/catalog/rawls/RawlsService.java
fboulnois: I added a ticket for this.

service/src/main/java/bio/terra/catalog/service/DatasetService.java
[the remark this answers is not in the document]
fboulnois: Okay, I've removed this.

service/src/test/java/bio/terra/catalog/service/DatasetServiceTest.java
fboulnois: Since I created a mock external dataset, I also created a mock `externalSourceId`. The regular `sourceId` could work as long as the mock external dataset also contained that `sourceId`. It would be confusing to me if I was looking in the debugger and I saw a mock external dataset with just the  regular `sourceId`, though (in fact this happened to me initially).

service/src/test/java/bio/terra/catalog/service/DatasetServiceTest.java
fboulnois: It does work without reset, I've removed those.

service/src/test/java/bio/terra/catalog/service/DatasetServiceTest.java
fboulnois: Do you mean "not sure why these fields **aren't** static"? If so, should I make them all static?

service/src/test/java/bio/terra/catalog/service/DatasetServiceTest.java
fboulnois: I ended up removing this due to removing the test for external datasets.

service/src/test/java/bio/terra/catalog/service/DatasetServiceTest.java
fboulnois: I removed this test.

service/src/test/java/bio/terra/catalog/service/DatasetServiceTest.java
pshapiro4broad: > Do you mean "not sure why these fields aren't static"? If so, should I make them all static?

Yes, that's what I meant. If you could make them all static I think that would be better.

pshapiro4broad:
Before merging could you look at the two "code smells" reported by sonar? https://sonarcloud.io/project/issues?id=terra-data-catalog&pullRequest=114&resolved=false&types=CODE_SMELL

service/src/main/java/bio/terra/catalog/rawls/RawlsService.java
```diff
@@ -99,4 +103,44 @@ public SystemStatusSystems status() {
...
+      WorkspaceDetails workspaceDetailsSource =
+          workspacesApi(user).getWorkspaceById(workspaceIdSource, List.of()).getWorkspace();
+      WorkspaceName workspaceNameSource = getWorkspaceName(workspaceDetailsSource);
+
+      // build destination name
+      WorkspaceDetails workspaceDetailsDest =
+          workspacesApi(user).getWorkspaceById(workspaceIdDest, List.of()).getWorkspace();
+      WorkspaceName workspaceNameDest = getWorkspaceName(workspaceDetailsDest);
+
+      // possible bug: empty entityType and entityNames copies all entities
```
s-rubenstein, line 130: I wonder if they are meant to be optional parameters, and the default is all?

s-rubenstein (approved):
LGTM

PR #115, opened 2022-08-03
DC-424: verify metadata on create/update

tcjiang9:
verify the metadata is a json object


service/src/main/java/bio/terra/catalog/service/DatasetService.java
```diff
@@ -229,6 +234,18 @@ public DatasetPreviewTablesResponse listDatasetPreviewTables(
     return new DatasetPreviewTablesResponse().tables(tableMetadataList);
   }
 
+  private void validateSchema(String metaData) {
```
pshapiro4broad, line 237: I think `validateMetadata()` would be a better name, as this is not validating a schema (or using a schema to validate the metadata).

pshapiro4broad, line 237: To be consistent with other code in our repo, this should be spelled `metadata`
```suggestion
  private void validateSchema(String metadata) {
```

service/src/main/java/bio/terra/catalog/service/DatasetService.java
```diff
@@ -229,6 +234,18 @@ public DatasetPreviewTablesResponse listDatasetPreviewTables(
...
+    JsonNode node;
+    try {
+      node = objectMapper.readValue(metaData, JsonNode.class);
+    } catch (JsonProcessingException e) {
+      throw new IllegalMetadataException(e);
+    }
+    if (node.getNodeType() != JsonNodeType.OBJECT) {
+      throw new BadRequestException("catalogEntry/metadata must be a json object");
+    }
```
pshapiro4broad, line 246: I think it's better to initialize a variable at its declaration point, if possible. Moving the `if` inside the `try` would allow you do do that.
```suggestion
    try {
      JsonNode node = objectMapper.readValue(metaData, JsonNode.class);
      if (node.getNodeType() != JsonNodeType.OBJECT) {
        throw new BadRequestException("catalogEntry/metadata must be a json object");
      }
    } catch (JsonProcessingException e) {
      throw new IllegalMetadataException(e);
    }
```

service/src/main/java/bio/terra/catalog/service/DatasetService.java
```diff
@@ -229,6 +234,18 @@ public DatasetPreviewTablesResponse listDatasetPreviewTables(
     return new DatasetPreviewTablesResponse().tables(tableMetadataList);
   }
 
+  private void validateSchema(String metaData) {
+    JsonNode node;
+    try {
+      node = objectMapper.readValue(metaData, JsonNode.class);
+    } catch (JsonProcessingException e) {
+      throw new IllegalMetadataException(e);
```
pshapiro4broad, line 242: Unlike the other code that's calling `objectMapper.readValue()` (in `DatasetService.toJsonNode()`), this exception could easily be seen by a user, so it should have a user visible message along with the caught exception.

service/src/main/java/bio/terra/catalog/service/DatasetService.java
```diff
@@ -229,6 +234,18 @@ public DatasetPreviewTablesResponse listDatasetPreviewTables(
...
+    }
+    if (node.getNodeType() != JsonNodeType.OBJECT) {
```
pshapiro4broad, line 244: Instead of doing this check, you could instead call `readvalue()` with `ObjectNode.class` instead of `JsonNode.class`. Or, call `toJsonNode()` directly and add a user visible message to `toJsonNode()`.

service/src/test/java/bio/terra/catalog/service/DatasetServiceTest.java
```diff
@@ -193,7 +194,17 @@ void testUpdateMetadataWithInvalidUser() {
...
     mockDataset();
-    String metadata = "test metadata";
+    String metadata = """
+        {"test": "metadata"}""";
+    when(samService.hasGlobalAction(user, SamAction.UPDATE_ANY_METADATA)).thenReturn(true);
+    datasetService.updateMetadata(user, datasetId, metadata);
+    verify(datasetDao).update(dataset.withMetadata(metadata));
+  }
+
+  @Test(expected = BadRequestException.class)
```
pshapiro4broad, line 204: I don't think JUnit's `expected` is a good way to detect an expected exception, since there's no way to verify which statement produced the exception. Instead I'd assert on the `updateMetadata` call.
```
    assertThrows(BadRequestException.class, () ->  datasetService.updateMetadata(user, datasetId, metadata));
```

service/src/test/java/bio/terra/catalog/service/DatasetServiceTest.java
```diff
@@ -193,7 +194,17 @@ void testUpdateMetadataWithInvalidUser() {
...
+  void testUpdateMetadataInvalidInput() {
+    mockDataset();
+    String metadata = "metadata must be json object";
     when(samService.hasGlobalAction(user, SamAction.UPDATE_ANY_METADATA)).thenReturn(true);
     datasetService.updateMetadata(user, datasetId, metadata);
     verify(datasetDao).update(dataset.withMetadata(metadata));
```
pshapiro4broad, line 210: Does this work? I'd expect that the dao is never called. In fact you may want to assert that using `verifyNoInteractions()`.

service/src/test/java/bio/terra/catalog/service/DatasetServiceTest.java
```diff
@@ -209,7 +220,8 @@ void testCreateDatasetWithInvalidUser() {
 
   @Test
   void testCreateDataset() {
-    String metadata = "test metadata";
+    String metadata = """
```
pshapiro4broad, line 223: May as well make a string constant for the duplicate metadata used in these tests.

pshapiro4broad (changes requested):
I have a few changes to suggest, let me know what you think.

Also with this change do we still need need `DatasetServiceTest.listDatasetsIllegalMetadata()`? Is there a way for illegal data to be stored in the database that would trigger an error on read?

fboulnois (commented):
I agree with @pshapiro4broad 's comments. I'll take another look when these are resolved.

service/src/main/java/bio/terra/catalog/service/DatasetService.java
tcjiang9: I think having the check makes it more clear what the intention of the code is, as opposed to one lining it

service/src/main/java/bio/terra/catalog/service/DatasetService.java
s-rubenstein: TOL: I might actually call it: `validateStringIsJsonObject`. I could see us using this for other inputs we want to make sure are JSON objects, and it makes it very clear what it does. That said, I think that's entirely up to personal discretion, and could go either way.

Making it be metadata specific might make someone think that this is where the schema validation is happening - is that the intended direction of this fn?

service/src/main/java/bio/terra/catalog/service/DatasetService.java
s-rubenstein: If you used the name `validateStringIsJsonObject`, I would probably move to @pshapiro4broad's suggestion. Otherwise I agree with @tcjiang9, it is a little more clear this way without strong understanding of the internals of the library.

However, I might consider onelining it, and making the body of the try:
```
if (objectMapper.readValue(metadata, JsonNode.class).getNodeType() != JsonNodeType.OBJECT) {
  throw new BadRequestException("catalogEntry/metadata must be a json object");
}
```

fboulnois (approved):
latest changes lgtm

s-rubenstein (approved):
LGTM, but I think Phil will need to re-review before you can merge because of the way Request Changes works.

service/src/main/java/bio/terra/catalog/service/DatasetService.java
pshapiro4broad: My thinking was that you should use the same API that's used by the code that retrieves the contents of the metadata and sends it back to the user. So if that code needs `ObjectNode`, this code should use it too. What if `getNodeType()` returns `JsonNodeType.OBJECT` but `readValue()` to `ObjectNode` fails?

service/src/test/java/bio/terra/catalog/service/DatasetServiceTest.java
```diff
@@ -187,29 +189,36 @@ void testGetMetadataUsingTdrPermission() {
   void testUpdateMetadataWithInvalidUser() {
     mockDataset();
     assertThrows(
-        UnauthorizedException.class, () -> datasetService.updateMetadata(user, datasetId, "test"));
+        UnauthorizedException.class, () -> datasetService.updateMetadata(user, datasetId, metadata));
```
pshapiro4broad, line 192: `metadata` should be declared `static final` and should be uppercase

service/src/test/java/bio/terra/catalog/service/DatasetServiceTest.java
```diff
@@ -187,29 +189,36 @@ void testGetMetadataUsingTdrPermission() {
...
     datasetService.updateMetadata(user, datasetId, metadata);
     verify(datasetDao).update(dataset.withMetadata(metadata));
   }
 
+  @Test
+  void testUpdateMetadataInvalidInput() {
+    String metadata = "metadata must be json object";
+    assertThrows(
+        BadRequestException.class, () -> datasetService.updateMetadata(user, datasetId, metadata));
+    verifyNoInteractions(samService);
```
pshapiro4broad, line 208: I don't think this assert belongs here, since the correctness of the code doesn't depend on if SAM is used or not. You can imagine a version of `updateMetadata()` which checked permission first before checking for valid JSON. That's different than `verifyNoInteractions(datasetDao)`, since we want to make sure that no database changes are made if the json is invalid.

I see that the DAO is used for both `retrieve()` and `update()`, a more correct assert would only affect `update()`. Something like `verify(datasetDao, never()).update();` or `then(datasetDao).should(never()).update();`.

service/src/main/java/bio/terra/catalog/service/DatasetService.java
```diff
@@ -217,6 +221,7 @@ public DatasetId createDataset(
       StorageSystem storageSystem,
       String storageSourceId,
       String metadata) {
+    validateMetadata(metadata);
     var dataset = new Dataset(storageSourceId, storageSystem, metadata);
```
pshapiro4broad, line 225: It occurs to me that both validation and conversion from String to ObjectNode are really `Dataset` operations. Dataset itself should disallow creation with invalid JSON metadata and support a way to get the content as `ObjectNode`. But I don't see a way to do that without making our `ObjectMapper` instance a global variable.

pshapiro4broad (approved):
I'm unblocking the review although I have a few more comments that need to be addressed before merging.

service/src/main/java/bio/terra/catalog/service/DatasetService.java
```diff
@@ -229,6 +234,16 @@ public DatasetPreviewTablesResponse listDatasetPreviewTables(
     return new DatasetPreviewTablesResponse().tables(tableMetadataList);
   }
 
+  private void validateMetadata(String metadata) {
+    try {
+      if (metadata == null || objectMapper.readValue(metadata, JsonNode.class).getNodeType() != JsonNodeType.OBJECT) {
```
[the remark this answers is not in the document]
tcjiang9: I added this because a unit test was passing a null value, it's not technically possible through using our API, so I can remove it.

service/src/test/java/bio/terra/catalog/service/DatasetServiceTest.java
```diff
@@ -193,29 +180,36 @@ void testGetMetadataUsingTdrPermission() {
...
     mockDataset();
-    String metadata = "test metadata";
     when(samService.hasGlobalAction(user, SamAction.UPDATE_ANY_METADATA)).thenReturn(true);
     datasetService.updateMetadata(user, datasetId, metadata);
     verify(datasetDao).update(dataset.withMetadata(metadata));
   }
 
+  @Test
+  void testUpdateMetadataInvalidInput() {
+    String metadata = "metadata must be json object";
```
pshapiro4broad, line 197: In general it's not good to have a variable shadow another variable in an outer scope. To avoid this, maybe use a different name for the local variable, like `badMetadata` or `notJson`.

pshapiro4broad (approved):

PR #116, opened 2022-08-04
DC-347: Preview data for workspaces

Abulu123:
Data preview seems to be working and all tests are complete!

### DIFF
--- a/service/src/main/java/bio/terra/catalog/rawls/RawlsService.java
+++ b/service/src/main/java/bio/terra/catalog/rawls/RawlsService.java
@@ -9,9 +9,9 @@ import bio.terra.rawls.api.StatusApi;
 import bio.terra.rawls.api.WorkspacesApi;
 import bio.terra.rawls.client.ApiClient;
 import bio.terra.rawls.client.ApiException;
+import bio.terra.rawls.model.EntityCopyDefinition;
 import bio.terra.rawls.model.EntityQueryResponse;
 import bio.terra.rawls.model.EntityTypeMetadata;
-import bio.terra.rawls.model.EntityCopyDefinition;
 import bio.terra.rawls.model.WorkspaceAccessLevel;
 import bio.terra.rawls.model.WorkspaceDetails;
 import bio.terra.rawls.model.WorkspaceName;
@@ -138,10 +138,6 @@ public class RawlsService {
     } catch (ApiException e) {
       throw new RawlsException("Entity Metadata failed for workspace %s".formatted(workspaceId), e);
     }
-  }
-
-  EntitiesApi entitiesApi(AuthenticatedUserRequest user) {
-    return new EntitiesApi(getApiClient(user));
   }
 
   @VisibleForTesting
--- a/service/src/main/java/bio/terra/catalog/rawls/RawlsService.java
+++ b/service/src/main/java/bio/terra/catalog/rawls/RawlsService.java
@@ -4,11 +4,15 @@ import bio.terra.catalog.config.RawlsConfiguration;
 import bio.terra.catalog.model.SystemStatusSystems;
 import bio.terra.catalog.service.dataset.DatasetAccessLevel;
 import bio.terra.common.iam.AuthenticatedUserRequest;
+import bio.terra.rawls.api.EntitiesApi;
 import bio.terra.rawls.api.StatusApi;
 import bio.terra.rawls.api.WorkspacesApi;
 import bio.terra.rawls.client.ApiClient;
 import bio.terra.rawls.client.ApiException;
+import bio.terra.rawls.model.EntityQueryResponse;
+import bio.terra.rawls.model.EntityTypeMetadata;
 import bio.terra.rawls.model.WorkspaceAccessLevel;
+import bio.terra.rawls.model.WorkspaceResponse;
 import com.google.common.annotations.VisibleForTesting;
 import java.util.List;
 import java.util.Map;
@@ -79,6 +83,64 @@ public class RawlsService {
 
   WorkspacesApi workspacesApi(AuthenticatedUserRequest user) {
     return new WorkspacesApi(getApiClient(user));
+  }
+
+  //  public List<Entity> entityList(AuthenticatedUserRequest user, String storageSourceId)  {
+  //    WorkspaceResponse response = null;
+  //    try {
+  //      response = workspacesApi(user).getWorkspaceById(storageSourceId, List.of());
+  //      var entities = entitiesApi(user).listEntities(response.getWorkspace().getNamespace(),
+  // response.getWorkspace().getName(), Sample);
+  //      return entities;
+  //    } catch (ApiException e) {
+  //      throw new RuntimeException(e);
+  //    }
+  //  }
+
+  public EntityQueryResponse entityQuery(
+      AuthenticatedUserRequest user, String storageSourceId, String tableName) {
+    WorkspaceResponse response;
+    try {
+      response = workspacesApi(user).getWorkspaceById(storageSourceId, List.of());
+      return entitiesApi(user)
+          .entityQuery(
+              response.getWorkspace().getNamespace(),
+              response.getWorkspace().getName(),
+              tableName,
+              null,
+              null,
+              null,
+              null,
+              null,
+              List.of(),
+              null,
+              null);
+    } catch (ApiException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  public Map<String, EntityTypeMetadata> entityMetadata(
+      AuthenticatedUserRequest user, String storageSourceId) {
+    WorkspaceResponse response;
+    try {
+      response = workspacesApi(user).getWorkspaceById(storageSourceId, List.of());
+      var entities =
+          entitiesApi(user)
+              .entityTypeMetadata(
+                  response.getWorkspace().getNamespace(),
+                  response.getWorkspace().getName(),
+                  true,
+                  null);
+      //noinspection unchecked
+      return (Map<String, EntityTypeMetadata>) entities;
+    } catch (ApiException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  EntitiesApi entitiesApi(AuthenticatedUserRequest user) {
+    return new EntitiesApi(getApiClient(user));
   }
 
   @VisibleForTesting
--- a/service/src/main/java/bio/terra/catalog/rawls/RawlsService.java
+++ b/service/src/main/java/bio/terra/catalog/rawls/RawlsService.java
@@ -85,23 +85,10 @@ public class RawlsService {
     return new WorkspacesApi(getApiClient(user));
   }
 
-  //  public List<Entity> entityList(AuthenticatedUserRequest user, String storageSourceId)  {
-  //    WorkspaceResponse response = null;
-  //    try {
-  //      response = workspacesApi(user).getWorkspaceById(storageSourceId, List.of());
-  //      var entities = entitiesApi(user).listEntities(response.getWorkspace().getNamespace(),
-  // response.getWorkspace().getName(), Sample);
-  //      return entities;
-  //    } catch (ApiException e) {
-  //      throw new RuntimeException(e);
-  //    }
-  //  }
-
   public EntityQueryResponse entityQuery(
-      AuthenticatedUserRequest user, String storageSourceId, String tableName) {
-    WorkspaceResponse response;
+      AuthenticatedUserRequest user, String workspaceId, String tableName) {
     try {
-      response = workspacesApi(user).getWorkspaceById(storageSourceId, List.of());
+      WorkspaceResponse response = workspacesApi(user).getWorkspaceById(workspaceId, List.of());
       return entitiesApi(user)
           .entityQuery(
               response.getWorkspace().getNamespace(),
@@ -116,15 +103,14 @@ public class RawlsService {
               null,
               null);
     } catch (ApiException e) {
-      throw new RawlsException("Entity Query Failed", e);
+      throw new RawlsException("Entity Query failed for workspace %s".formatted(workspaceId), e);
     }
   }
 
   public Map<String, EntityTypeMetadata> entityMetadata(
-      AuthenticatedUserRequest user, String storageSourceId) {
-    WorkspaceResponse response;
+      AuthenticatedUserRequest user, String workspaceId) {
     try {
-      response = workspacesApi(user).getWorkspaceById(storageSourceId, List.of());
+      WorkspaceResponse response = workspacesApi(user).getWorkspaceById(workspaceId, List.of());
       var entities =
           entitiesApi(user)
               .entityTypeMetadata(
@@ -147,7 +133,7 @@ public class RawlsService {
                             .attributeNames((List<String>) value.get("attributeNames"));
                       }));
     } catch (ApiException e) {
-      throw new RawlsException("Entity Metadata failed", e);
+      throw new RawlsException("Entity Metadata failed for workspace %s".formatted(workspaceId), e);
     }
   }
 
--- a/service/src/main/java/bio/terra/catalog/rawls/RawlsService.java
+++ b/service/src/main/java/bio/terra/catalog/rawls/RawlsService.java
@@ -102,6 +102,7 @@ public class RawlsService {
               null,
               null,
               null,
+              null,
               List.of(),
               null,
               null);
@@ -114,27 +115,12 @@ public class RawlsService {
       AuthenticatedUserRequest user, String workspaceId) {
     try {
       WorkspaceResponse response = workspacesApi(user).getWorkspaceById(workspaceId, List.of());
-      var entities =
-          entitiesApi(user)
-              .entityTypeMetadata(
-                  response.getWorkspace().getNamespace(),
-                  response.getWorkspace().getName(),
-                  true,
-                  null);
-      //noinspection unchecked
-      return ((Map<String, Map<String, Object>>) entities)
-          .entrySet().stream()
-              .collect(
-                  Collectors.toMap(
-                      Map.Entry::getKey,
-                      entry -> {
-                        Map<String, Object> value = entry.getValue();
-                        //noinspection unchecked
-                        return new EntityTypeMetadata()
-                            .idName((String) value.get("idName"))
-                            .count((Integer) value.get("count"))
-                            .attributeNames((List<String>) value.get("attributeNames"));
-                      }));
+      return entitiesApi(user)
+          .entityTypeMetadata(
+              response.getWorkspace().getNamespace(),
+              response.getWorkspace().getName(),
+              true,
+              null);
     } catch (ApiException e) {
       throw new RawlsException("Entity Metadata failed for workspace %s".formatted(workspaceId), e);
     }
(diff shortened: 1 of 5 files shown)


service/src/test/java/bio/terra/catalog/service/DatasetServiceTest.java
```diff
@@ -278,4 +298,27 @@ void getDatasetPreviewTable() {
...
+    when(rawlsService.entityMetadata(user, tdrDataset.storageSourceId())).thenReturn(map);
+    when(rawlsService.entityQuery(user, tdrDataset.storageSourceId(), tableName))
+        .thenReturn(new EntityQueryResponse().results(List.of()));
+    DatasetPreviewTable datasetPreviewTable =
+        datasetService.getDatasetPreview(user, tdrDataset.id(), tableName);
+    assertThat(datasetPreviewTable.getRows(), empty());
+    assertThat(datasetPreviewTable.getColumns(), hasSize(4));
+    assertThat(
+        datasetPreviewTable.getColumns().get(0),
+        is(new bio.terra.catalog.model.ColumnModel().name("idName")));
```
Abulu123, line 322: RawlsService does not return the data type for the columns. Our api returns whether or not the column is an array. Is this even something we want to implement?

service/src/main/java/bio/terra/catalog/rawls/RawlsService.java
```diff
@@ -81,6 +85,76 @@ WorkspacesApi workspacesApi(AuthenticatedUserRequest user) {
     return new WorkspacesApi(getApiClient(user));
   }
 
+  //  public List<Entity> entityList(AuthenticatedUserRequest user, String storageSourceId)  {
```
s-rubenstein, line 88: If we don't need this, can we remove the commented out code?

service/src/main/java/bio/terra/catalog/rawls/RawlsService.java
```diff
@@ -81,6 +85,76 @@ WorkspacesApi workspacesApi(AuthenticatedUserRequest user) {
...
+              null,
+              null,
+              null,
+              null,
+              null,
+              List.of(),
+              null,
+              null);
+    } catch (ApiException e) {
+      throw new RawlsException("Entity Query Failed", e);
```
s-rubenstein, line 119: I would add the workspace to this exception message: `"Entity Query Failed for workspace %s", storageSourceId`

service/src/main/java/bio/terra/catalog/rawls/RawlsService.java
```diff
@@ -81,6 +85,76 @@ WorkspacesApi workspacesApi(AuthenticatedUserRequest user) {
...
+                      entry -> {
+                        Map<String, Object> value = entry.getValue();
+                        //noinspection unchecked
+                        return new EntityTypeMetadata()
+                            .idName((String) value.get("idName"))
+                            .count((Integer) value.get("count"))
+                            .attributeNames((List<String>) value.get("attributeNames"));
+                      }));
+    } catch (ApiException e) {
+      throw new RawlsException("Entity Metadata failed", e);
```
s-rubenstein, line 150: I would also add the workspace ID for this error message.

service/src/main/java/bio/terra/catalog/service/DatasetService.java
```diff
@@ -147,15 +150,18 @@ private List<TableMetadata> generateDatasetTables(
     return switch (dataset.storageSystem()) {
       case TERRA_DATA_REPO -> convertDatarepoTablesToCatalogTables(
           datarepoService.getPreviewTables(user, dataset.storageSourceId()).getTables());
-      case TERRA_WORKSPACE, EXTERNAL -> List.of();
+      case TERRA_WORKSPACE -> convertRawlsTablesToCatalogTables(
```
s-rubenstein, line 153: I'm wondering if the conversion methods should go into the service layer, so that the datarepo and rawls services return the same shape of data. That might make more sense, because it would mean that the `DatasetService` wouldn't need to know about specific service API shape.

service/src/test/java/bio/terra/catalog/rawls/RawlsServiceTest.java
```diff
@@ -109,4 +116,60 @@ void userHasActionException() throws Exception {
...
+    WorkspaceResponse response =
+        new WorkspaceResponse().workspace(new WorkspaceDetails().name(name).namespace(namespace));
+    when(workspacesApi.getWorkspaceById(id, List.of())).thenReturn(response);
+    Map<String, EntityTypeMetadata> queryResponse = new HashMap<>();
+    when(entitiesApi.entityTypeMetadata(namespace, name, true, null)).thenReturn(queryResponse);
+    assertThat(rawlsService.entityMetadata(user, id), is(queryResponse));
+  }
+
+  @Test
+  void entityMetadataExecption() throws Exception {
```
s-rubenstein, line 165: Nit: Exception is spelled wrong here.

service/src/test/java/bio/terra/catalog/service/DatasetServiceTest.java
```diff
@@ -2,6 +2,7 @@
 
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.*;
+import static org.hamcrest.Matchers.is;
```
s-rubenstein, line 5: It looks like we already do a `.*` import above? Could we use `Matchers.is` here, rather than add the redundant import?

service/src/main/java/bio/terra/catalog/service/DatasetService.java
pshapiro4broad: I was thinking the same thing, that the conversion code for both TDR and Rawls should be pushed into the lower service layers. But maybe that could wait for a future refactor? This code is (over) due for that kind of work.

service/src/test/java/bio/terra/catalog/service/DatasetServiceTest.java
pshapiro4broad: Yes, in general we don't want to use wildcard (`*`) imports.

service/src/test/java/bio/terra/catalog/service/DatasetServiceTest.java
pshapiro4broad: I think it's OK. We may want to revisit this and remove the type information from our APIs, if the UI doesn't need this information to render the preview.

service/src/main/java/bio/terra/catalog/rawls/RawlsService.java
pshapiro4broad: It couldn't hurt although the workspace ID is unknown to most users. It may be helpful for developers though.

Also the capitalization here doesn't match other error messages in this code. It looks like most errors are sentence fragments that begin with a capital letter.

service/src/main/java/bio/terra/catalog/service/DatasetService.java
```diff
@@ -181,6 +187,39 @@ private DatasetPreviewTable generateDatarepoTable(
...
+                .entityQuery(user, dataset.storageSourceId(), tableName)
+                .getResults()
+                .stream()
+                .map(Entity -> convertEntityToRow(Entity, obj.getIdName()))
+                .toList());
+  }
+
+  private Object convertEntityToRow(Entity entity, String idName) {
+    @SuppressWarnings("unchecked")
+    Map<String, String> att = (Map<String, String>) entity.getAttributes();
```
pshapiro4broad, line 208: This should also be fixable by updating the Rawls swagger.

service/src/main/java/bio/terra/catalog/rawls/RawlsService.java
```diff
@@ -81,6 +85,76 @@ WorkspacesApi workspacesApi(AuthenticatedUserRequest user) {
...
+  //      var entities = entitiesApi(user).listEntities(response.getWorkspace().getNamespace(),
+  // response.getWorkspace().getName(), Sample);
+  //      return entities;
+  //    } catch (ApiException e) {
+  //      throw new RuntimeException(e);
+  //    }
+  //  }
+
+  public EntityQueryResponse entityQuery(
+      AuthenticatedUserRequest user, String storageSourceId, String tableName) {
```
pshapiro4broad, line 101: `storageSourceId` is the name of this concept in the caller. In this code, it's `workspaceId`.
```suggestion
      AuthenticatedUserRequest user, String workspaceId, String tableName) {
```

service/src/main/java/bio/terra/catalog/rawls/RawlsService.java
```diff
@@ -81,6 +85,76 @@ WorkspacesApi workspacesApi(AuthenticatedUserRequest user) {
...
+    } catch (ApiException e) {
+      throw new RawlsException("Entity Query Failed", e);
+    }
+  }
+
+  public Map<String, EntityTypeMetadata> entityMetadata(
+      AuthenticatedUserRequest user, String storageSourceId) {
+    WorkspaceResponse response;
+    try {
+      response = workspacesApi(user).getWorkspaceById(storageSourceId, List.of());
```
pshapiro4broad, line 127: It looks like this declaration can be moved inside the `try`
```suggestion
    try {
      WorkspaceResponse response = workspacesApi(user).getWorkspaceById(storageSourceId, List.of());
```

service/src/main/java/bio/terra/catalog/rawls/RawlsService.java
```diff
@@ -81,6 +85,76 @@ WorkspacesApi workspacesApi(AuthenticatedUserRequest user) {
...
+      var entities =
+          entitiesApi(user)
+              .entityTypeMetadata(
+                  response.getWorkspace().getNamespace(),
+                  response.getWorkspace().getName(),
+                  true,
+                  null);
+      //noinspection unchecked
+      return ((Map<String, Map<String, Object>>) entities)
```
pshapiro4broad, line 136: Most of this code should go away once the Rawls swagger is updated.

service/src/main/java/bio/terra/catalog/rawls/RawlsService.java
```diff
@@ -81,6 +85,76 @@ WorkspacesApi workspacesApi(AuthenticatedUserRequest user) {
...
+  //    } catch (ApiException e) {
+  //      throw new RuntimeException(e);
+  //    }
+  //  }
+
+  public EntityQueryResponse entityQuery(
+      AuthenticatedUserRequest user, String storageSourceId, String tableName) {
+    WorkspaceResponse response;
+    try {
+      response = workspacesApi(user).getWorkspaceById(storageSourceId, List.of());
```
pshapiro4broad, line 104: It looks like this declaration can be moved inside the `try`
```suggestion
    try {
      WorkspaceResponse response = workspacesApi(user).getWorkspaceById(storageSourceId, List.of());
```

service/src/main/java/bio/terra/catalog/service/DatasetService.java
```diff
@@ -181,6 +187,39 @@ private DatasetPreviewTable generateDatarepoTable(
                 .getResult());
   }
 
+  private DatasetPreviewTable generateRawlsTable(
+      AuthenticatedUserRequest user, Dataset dataset, String tableName) {
+    Map<String, EntityTypeMetadata> ent =
+        rawlsService.entityMetadata(user, dataset.storageSourceId());
+    EntityTypeMetadata obj = ent.get(tableName);
```
pshapiro4broad, line 194: I think we should have better names here than `ent` and `obj`. In general we try to use complete words, not abbreviations, for names. Maybe `entityMetadata` for `ent` and `tableMetadata` for `obj`?

service/src/main/java/bio/terra/catalog/service/DatasetService.java
```diff
@@ -181,6 +187,39 @@ private DatasetPreviewTable generateDatarepoTable(
...
+
+  private Object convertEntityToRow(Entity entity, String idName) {
+    @SuppressWarnings("unchecked")
+    Map<String, String> att = (Map<String, String>) entity.getAttributes();
+    Map<String, String> rows = new HashMap<>(att);
+    rows.put(idName, entity.getName());
+    return rows;
+  }
+
+  private static List<ColumnModel> convertEntityToColumn(EntityTypeMetadata entity) {
```
pshapiro4broad, line 214: This is actually converting the entity type metadata (which is the table metadata) to columns. So maybe the name should be `convertTableMetadataToColumns`?

service/src/main/java/bio/terra/catalog/service/DatasetService.java
```diff
@@ -181,6 +187,39 @@ private DatasetPreviewTable generateDatarepoTable(
...
+    List<ColumnModel> model = new ArrayList<>();
```
pshapiro4broad, line 215: Instead of `model` I would call this `columns`, since it is a list of columns.

service/src/main/java/bio/terra/catalog/service/DatasetService.java
```diff
@@ -181,6 +187,39 @@ private DatasetPreviewTable generateDatarepoTable(
...
+        rawlsService.entityMetadata(user, dataset.storageSourceId());
+    EntityTypeMetadata obj = ent.get(tableName);
+    return new DatasetPreviewTable()
+        .columns(convertEntityToColumn(obj))
+        .rows(
+            rawlsService
+                .entityQuery(user, dataset.storageSourceId(), tableName)
+                .getResults()
+                .stream()
+                .map(Entity -> convertEntityToRow(Entity, obj.getIdName()))
```
pshapiro4broad, line 202: In general variable names start with a lower case letter, so this should be `entity` not `Entity`
```suggestion
                .map(entity -> convertEntityToRow(entity, obj.getIdName()))
```

service/src/main/java/bio/terra/catalog/service/DatasetService.java
```diff
@@ -240,6 +279,13 @@ private static List<TableMetadata> convertDatarepoTablesToCatalogTables(
         .toList();
   }
 
+  private static List<TableMetadata> convertRawlsTablesToCatalogTables(
+      Map<String, EntityTypeMetadata> entityTables) {
+    return entityTables.keySet().stream()
+        .map(entityTypeMetadata -> new TableMetadata().name(entityTypeMetadata).hasData(true))
```
pshapiro4broad, line 285: I don't think `hasData(true)` is correct. This should this be something like `hasData(entityTables.get(entityTypeMetadata).getCount() != 0)`.

Although if you do this, then this should use `entrySet().stream()` instead of `keySet().stream()`. Maybe

```
    return entityTables.entrySet().stream()
        .map(entry -> new TableMetadata().name(entry.getKey()).hasData(entry.getValue().getCount() != 0))
        .toList();
```
Also there should be a unit test for this case.

service/src/test/java/bio/terra/catalog/rawls/RawlsServiceTest.java
```diff
@@ -109,4 +116,60 @@ void userHasActionException() throws Exception {
...
+    String id = "abc";
+    String name = "name";
+    String namespace = "namespace";
+    String tableName = "table";
+    WorkspaceResponse response =
+        new WorkspaceResponse().workspace(new WorkspaceDetails().name(name).namespace(namespace));
+    when(workspacesApi.getWorkspaceById(id, List.of())).thenThrow(new ApiException());
+    when(entitiesApi.entityQuery(
+            namespace, name, tableName, null, null, null, null, null, List.of(), null, null))
+        .thenThrow(new ApiException());
```
pshapiro4broad, line 147: I'm a bit surprised that this passes with strict mocks enabled, since I don't see how both mocks are called. Once `getWorkspaceById()` throws, `entitiesApi.entityQuery()` won't be called.

service/src/test/java/bio/terra/catalog/rawls/RawlsServiceTest.java
```diff
@@ -109,4 +116,60 @@ void userHasActionException() throws Exception {
...
+
+  @Test
+  void entityMetadataExecption() throws Exception {
+    String id = "abc";
+    String name = "name";
+    String namespace = "namespace";
+    WorkspaceResponse response =
+        new WorkspaceResponse().workspace(new WorkspaceDetails().name(name).namespace(namespace));
+    when(workspacesApi.getWorkspaceById(id, List.of())).thenThrow(new ApiException());
+    when(entitiesApi.entityTypeMetadata(namespace, name, true, null)).thenThrow(new ApiException());
```
pshapiro4broad, line 172: Same comment as above; if one API throws then the next won't be called.

pshapiro4broad (commented):
A few comments. Also, be sure to check out the "code smell" report from sonarqube.

service/src/main/java/bio/terra/catalog/service/DatasetService.java
pshapiro4broad: Which will be done in a Rawls PR, not in this PR. If it's not in before this is merged, it's ok to keep the cast in for now.

service/src/test/java/bio/terra/catalog/service/DatasetServiceTest.java
pshapiro4broad: There's an IDE setting that should prevent this from happening, I can show you how to enable it if you haven't yet.

service/src/main/java/bio/terra/catalog/rawls/RawlsService.java
```diff
@@ -81,6 +85,76 @@ WorkspacesApi workspacesApi(AuthenticatedUserRequest user) {
...
+                            .attributeNames((List<String>) value.get("attributeNames"));
+                      }));
+    } catch (ApiException e) {
+      throw new RawlsException("Entity Metadata failed", e);
+    }
+  }
+
+  EntitiesApi entitiesApi(AuthenticatedUserRequest user) {
+    return new EntitiesApi(getApiClient(user));
+  }
```
fboulnois, line 156: When you rebase you should be able to pick up this change from my code.

service/src/main/java/bio/terra/catalog/service/DatasetService.java
fboulnois: Yes, I think it's worth making a ticket, but not blocking this work.

service/src/main/java/bio/terra/catalog/service/DatasetService.java
fboulnois, line 208:
```suggestion
    Map<String, String> attributeMap = (Map<String, String>) entity.getAttributes();
```

service/src/test/java/bio/terra/catalog/rawls/RawlsServiceTest.java
```diff
@@ -44,17 +48,20 @@ class RawlsServiceTest {
 
   @Mock private WorkspacesApi workspacesApi;
 
+  @Mock private EntitiesApi entitiesApi;
```
fboulnois, line 51: When you rebase you should be able to pick up this change from the code.

service/src/main/java/bio/terra/catalog/rawls/RawlsService.java
Abulu123: sounds good

service/src/main/java/bio/terra/catalog/rawls/RawlsService.java
Abulu123: will do

pshapiro4broad (approved):
Looks good. I think we'll want to revisit the dataset tests at some point in the future to better test the preview results. Making the catalog preview result typed in swagger will make that easier.

s-rubenstein (approved):
LGTM (Looks good to me)

PR #117, opened 2022-08-18
DC-509: integration tests for preview data

Abulu123:


integration/src/main/java/scripts/client/RawlsClient.java
```diff
@@ -116,9 +133,28 @@ public WorkspaceDetails createTestWorkspace() throws ApiException {
...
+    ingestData(workspaceDetails.getNamespace(), workspaceDetails.getName());
     return workspaceDetails;
   }
 
+  private void ingestData(String namespace, String name) throws ApiException {
+    for (int i = 1; i <= 15; i++) {
+      Entity entity = new Entity().entityType("sample").name("sample" + i);
+      entity.setAttributes(
+          Map.of(
+              //          "files",
```
pshapiro4broad, line 145: How do you plan to handle array data in Rawls? If storing and retrieving it works fine, the question then is if you need to property support the `arrayOf()` property in `ColumnModel`. To find that out, you may have to look at the UI code to see if it relies on this value.

If it is needed, the question then is how can you property populate this information in `ColumnModel`. One approach would be to read the first row in the table and use the type of the attribute to generate this value.

service/src/test/java/bio/terra/catalog/service/DatasetServiceTest.java
```diff
@@ -300,15 +300,15 @@ void getDatasetPreviewTableRepo() {
                                     new ColumnModel()
                                         .datatype(TableDataType.INTEGER)
                                         .name("column a"))))));
-    when(datarepoService.getPreviewTable(user, tdrDataset.storageSourceId(), tableName))
+    when(datarepoService.getPreviewTable(user, tdrDataset.storageSourceId(), tableName, 30))
```
pshapiro4broad, line 303: Using `30` here makes this test dependent on the constant `MAX_ROWS` without referring to it directly. Instead this should either refer to the constant (by making the constant visible to this code by removing `private`) or by changing the mock argument list to use argument matchers. Using an argument matcher would look something like:
```suggestion
    when(datarepoService.getPreviewTable(eq(user), eq(tdrDataset.storageSourceId()), eq(tableName), anyInt()))
```

pshapiro4broad (approved):
Looks good!

tcjiang9 (approved):

PR #118, opened 2022-08-19
DC-330: Add export to workspace integration tests

fboulnois:


integration/src/main/java/scripts/testscripts/DatasetOperations.java
```diff
@@ -63,7 +66,8 @@ public void setup(List<TestUserSpecification> testUsers) throws Exception {
 
   private void setupWorkspace(TestUserSpecification user) throws Exception {
     rawlsClient = new RawlsClient(server, user);
-    workspaceDetails = rawlsClient.createTestWorkspace();
+    workspaceSource = rawlsClient.createTestWorkspace();
+    workspaceDest = rawlsClient.createTestWorkspace();
```
pshapiro4broad, line 70: Since this is only used inside `exportUserJourney()` I would create it there. Also when we test both TDR and workspace export we'd want to make a new destination workspace each time.

integration/src/main/java/scripts/testscripts/DatasetOperations.java
```diff
@@ -85,9 +88,42 @@ public void userJourney(TestUserSpecification testUser) throws Exception {
...
+
+    // Export workspace to workspace
+    var workspaceId = UUID.fromString(workspaceDest.getWorkspaceId());
+    datasetsApi.exportDataset(datasetId, workspaceId);
+
+    // Extract workspace entity names
+    Set<String> entitiesSource = rawlsClient.getWorkspaceEntities(workspaceSource);
+    Set<String> entitiesDest = rawlsClient.getWorkspaceEntities(workspaceDest);
+
+    assertEquals(entitiesSource, entitiesDest);
```
pshapiro4broad, line 118: What entities are in these sets? Right now I don't think `workspaceSource` contains any entities. Do you plan to wait for Omesi's PR to merge up, which does populate the workspace?

pshapiro4broad (approved):
The code looks fine, although I'm not sure the assert is a good test yet.

integration/src/main/java/scripts/testscripts/DatasetOperations.java
```diff
@@ -85,9 +89,37 @@ public void userJourney(TestUserSpecification testUser) throws Exception {
...
+      throws ApiException, bio.terra.rawls.client.ApiException {
+    // Create workspace dataset
+    var request =
+        new CreateDatasetRequest()
+            .catalogEntry(METADATA_1)
+            .storageSourceId(workspaceSource.getWorkspaceId())
+            .storageSystem(storageSystem);
+    datasetId = datasetsApi.createDataset(request).getId();
+
+    // Export workspace to workspace
```
tcjiang9, line 110: export workspace dataset to workspace would be more clear I think

tcjiang9 (approved):

PR #119, opened 2022-08-26
DC-523: Add empty json to avoid Content-Length errors

fboulnois:


common/openapi.yml
```diff
@@ -153,6 +153,11 @@ paths:
         - $ref: '#/components/parameters/Id'
         - $ref: '#/components/parameters/WorkspaceId'
       operationId: exportDataset
+      requestBody:
+        description: A empty json request body to avoid Content-Length errors
+        content:
+          application/json: { }
```
pshapiro4broad, line 159: If this does fix the problem, maybe we should move `workspaceId` to the body so it will always be present?

fboulnois: Thoughts on your earlier comment? https://github.com/DataBiosphere/terra-data-catalog/pull/113#discussion_r930521292

Abulu123 (commented):
looks good

Abulu123 (approved):

PR #120, opened 2022-08-26
DC-269: fix sonar errors

pshapiro4broad:
Our repo contains many sonar warnings / errors, mostly in code that was added before we started doing sonar scans. This PR fixes these warnings. The overall scan of the repo is here https://sonarcloud.io/summary/overall?id=terra-data-catalog

- fixed all smells / bugs / hotspots
- added some code coverage
- added sonarqube badges to readme for coverage and quality gate
- updated a few gradle plugin dependencies to current versions
- moved controller classes into `controller` package since `api` is used by the swagger generated server code


.github/workflows/build-and-test.yml
```diff
@@ -31,10 +31,10 @@ jobs:
...
         run: ./gradlew --build-cache build -x test
 
-      - name: Upload spotbugs results
-        uses: github/codeql-action/upload-sarif@main
-        with:
-          sarif_file: service/build/reports/spotbugs/main.sarif
+#      - name: Upload spotbugs results
+#        uses: github/codeql-action/upload-sarif@main
+#        with:
+#          sarif_file: service/build/reports/spotbugs/main.sarif
```
fboulnois, line 37: Should we completely remove this now?

buildSrc/build.gradle
```diff
@@ -15,7 +15,7 @@ dependencies {
     implementation 'com.google.cloud.tools.jib:com.google.cloud.tools.jib.gradle.plugin:3.2.1'
     implementation 'com.srcclr.gradle:com.srcclr.gradle.gradle.plugin:3.1.12'
     implementation 'de.undercouch.download:de.undercouch.download.gradle.plugin:5.1.0'
-    implementation 'gradle.plugin.com.github.spotbugs.snom:spotbugs-gradle-plugin:4.7.2'
+    implementation 'ru.vyarus:gradle-quality-plugin:4.7.0'
```
fboulnois, line 18: Are there any concerns with using this plugin over spotbugs? It seems like it's maintained by a single person as a personal project. The latest version is 4.8.0.

service/src/main/java/bio/terra/catalog/iam/SamService.java
```diff
@@ -83,7 +83,7 @@ public SystemStatusSystems status() {
       var result = new SystemStatusSystems().ok(samStatus.getOk());
       var samSystems = samStatus.getSystems();
       // Populate error message if Sam status is non-ok
-      if (!samStatus.getOk()) {
+      if (!Boolean.TRUE.equals(result.isOk())) {
```
fboulnois, line 86: This is verbose. Is there no better way to express this condition?

fboulnois (approved):
lgtm, a few questions and feedback:

.github/workflows/build-and-test.yml
pshapiro4broad: I forgot about this, I'll either remove it or fix it. Spotbugs does seem to catch things that sonarqube doesn't report. I seem to recall there was a problem getting it to use sarif output.

buildSrc/build.gradle
pshapiro4broad: I switched this because I couldn't figure out how to exclude generated code using the other plugin without changing the generated code package. It has the benefit of supporting a few different code analysis tools too. Right now we're running spotbugs but using the most lenient setting, switching it to a different setting would require excluding these files. But this is all separate from fixing sonar errors so I can back this part out and create a separate tech debt ticket for spotbugs.

service/src/main/java/bio/terra/catalog/iam/SamService.java
pshapiro4broad: I'm pretty sure I tried
```suggestion
      if (result.isOk() == null || !result.isOk()) {
```
but I can try it again. I'm not sure this is much better though.

pshapiro4broad: Oh, I could write
```suggestion
      if (Boolean.FALSE.equals(result.isOk())) {
```

buildSrc/build.gradle
pshapiro4broad: I reverted this change

PR #122, opened 2022-08-31
DC-527: Separate Trivy from other Github actions

fboulnois:
Trivy success: https://github.com/DataBiosphere/terra-data-catalog/actions/runs/2966902191

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -62,11 +62,6 @@ jobs:
             --image=${{ steps.image-name.outputs.name }} \
             -Djib.console=plain
 
-      - name: Run Trivy vulnerability scanner
-        uses: broadinstitute/dsp-appsec-trivy-action@v1
-        with:
-          image: ${{ steps.image-name.outputs.name }}
-
   source-clear:
     needs: [ build ]
     runs-on: ubuntu-latest
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -62,6 +62,19 @@ jobs:
             --image=${{ steps.image-name.outputs.name }} \
             -Djib.console=plain
 
+#  dispatch-trivy:
+#    needs: [ build ]
+#    runs-on: ubuntu-latest
+#
+#    if: github.event_name == 'pull_request'
+#
+#    steps:
+#      - name: Fire off Trivy action
+#        uses: broadinstitute/workflow-dispatch@v1
+#        with:
+#          workflow: Trivy
+#          token: ${{ secrets.BROADBOT_TOKEN }}
+
   source-clear:
     needs: [ build ]
     runs-on: ubuntu-latest
@@ -188,18 +201,6 @@ jobs:
           fields: workflow,message
           text: 'Build failed :sadpanda:'
           username: 'Data Explorer GitHub Action'
-
-  dispatch-trivy:
-    runs-on: ubuntu-latest
-
-    if: github.event_name == 'pull_request'
-
-    steps:
-      - name: Fire off Trivy action
-        uses: broadinstitute/workflow-dispatch@v1
-        with:
-          workflow: Trivy
-          token: ${{ secrets.BROADBOT_TOKEN }}
 
   dispatch-tag:
     needs: [ build, unit-tests-and-sonarqube, source-clear, integration-tests ]
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -189,6 +189,18 @@ jobs:
           text: 'Build failed :sadpanda:'
           username: 'Data Explorer GitHub Action'
 
+  dispatch-trivy:
+    runs-on: ubuntu-latest
+
+    if: github.event_name == 'pull_request'
+
+    steps:
+      - name: Fire off Trivy action
+        uses: broadinstitute/workflow-dispatch@v1
+        with:
+          workflow: Trivy
+          token: ${{ secrets.BROADBOT_TOKEN }}
+
   dispatch-tag:
     needs: [ build, unit-tests-and-sonarqube, source-clear, integration-tests ]
     runs-on: ubuntu-latest
--- a/.github/workflows/nightly-tests.yml
+++ b/.github/workflows/nightly-tests.yml
@@ -14,7 +14,7 @@ on:
           - dev
         default: 'perf'
   schedule:
-    - cron: "0 6 * * *" # runs at 6 AM UTC, 2 AM EST.
+    - cron: "0 6 * * *" # runs at 6 AM UTC, 1 AM EST.
 
 env:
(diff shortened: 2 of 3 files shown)


.github/workflows/nightly-tests.yml
```diff
@@ -14,7 +14,7 @@ on:
           - dev
         default: 'perf'
   schedule:
-    - cron: "0 6 * * *" # runs at 6 AM UTC, 1 AM EST.
+    - cron: "0 6 * * *" # runs at 6 AM UTC, 2 AM EST.
```
pshapiro4broad, line 17: 1am is correct for EST (UTC-05). 2am is EDT.

.github/workflows/trivy.yml
```diff
@@ -0,0 +1,57 @@
...
+
+      - name: Build all projects without running tests
+        run: ./gradlew --build-cache build -x test -x spotlessCheck
+
+      - name: Construct docker image name and tag
+        id: image-name
+        run: |
+          GITHUB_REPO=$(basename ${{ github.repository }})
+          GIT_SHORT_HASH=$(git rev-parse --short HEAD)
+          echo ::set-output name=name::${GITHUB_REPO}:${GIT_SHORT_HASH}
```
pshapiro4broad, line 29: I guess it doesn't hurt to do this but since the image isn't pushed to artifactory I don't think it matters what the image is called. You could use a fixed name for both the build output and the scanner input.

.github/workflows/trivy.yml
```diff
@@ -0,0 +1,57 @@
...
+
+      - name: Build image locally with jib
```
pshapiro4broad, line 31: Could this instead run the scan on the latest built image in artifactory or gcr? What is the benefit of building it here instead of getting it from there? In a way, I think scanning the published image is more important since it doesn't matter if it builds and scans if the published copy is invalid.

Maybe something like
```
        run: docker pull gcr.io/broad-dsp-gcr-public/terra-data-catalog
```

.github/workflows/trivy.yml
```diff
@@ -0,0 +1,57 @@
+name: Trivy Scanner
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 8 * * *" # runs at 8 AM UTC, 4 AM EST.
```
pshapiro4broad, line 6:
```suggestion
    - cron: "0 8 * * *" # runs at 8 AM UTC, 3 AM EST.
```
Any reason to not run this at 1AM like the nightly test? Is the worry that it will compete for GitHub runner resources?

pshapiro4broad (approved):
This is OK but I think we should be scanning a built image instead of building it before scanning.

.github/workflows/nightly-tests.yml
fboulnois: Maybe I'm missing something: https://www.google.com/search?q=6+am+utc+to+est

.github/workflows/trivy.yml
fboulnois: Yes, I think we should stagger the Github actions.

.github/workflows/nightly-tests.yml
pshapiro4broad: Interesting, although EST is still UTC-5 and EDT is UTC-4. When I use that link it mentions ET (aka Eastern Time) but that's something else.

We were so close to going DST all the time in 2023 but it looks like that's not going to happen anytime soon. https://thehill.com/homenews/house/3571007-permanent-daylight-saving-time-hits-brick-wall-in-house/

Anyways I still think UTC-5 is correct for EST. If you like you can remove the EST from the comment completely since UTC alone is sufficient and doesn't change twice a year.

pshapiro4broad: See https://en.wikipedia.org/wiki/Eastern_Time_Zone

.github/workflows/trivy.yml
pshapiro4broad: Why do you think we should do this? What improvements will we see if this is changed?

.github/workflows/trivy.yml
fboulnois: I went back to our production readiness documentation and for compliance and security it says: "Trivy is set up as a GitHub Action and runs against **each PR**". I need to rework this action.

.github/workflows/trivy.yml
fboulnois: I think we should stagger Github actions to avoid resource contention and to make debugging easier.

.github/workflows/trivy.yml
fboulnois: Additionally, I'm not sure how useful it is to scan existing images: this will not help us be proactive about upcoming issues, and will only tell us existing issues which might already be fixed.

.github/workflows/build-and-test.yml
```diff
@@ -62,10 +62,18 @@ jobs:
             --image=${{ steps.image-name.outputs.name }} \
             -Djib.console=plain
 
-      - name: Run Trivy vulnerability scanner
-        uses: broadinstitute/dsp-appsec-trivy-action@v1
-        with:
-          image: ${{ steps.image-name.outputs.name }}
+#  dispatch-trivy:
```
fboulnois, line 65: Because the Trivy.yml doesn't exist yet, this step will fail. Once this PR is merged, a follow-up PR will uncomment these.

.github/workflows/trivy.yml
fboulnois: I'm okay with keeping this as is to be consistent with our other pipelines.

.github/workflows/trivy.yml
```diff
@@ -0,0 +1,53 @@
+name: Trivy Scanner
+on: workflow_dispatch
```
fboulnois, line 2: This has manually been dispatched to ensure it works as expected.

.github/workflows/trivy.yml
pshapiro4broad: > I went back to our production readiness documentation and for compliance and security it says: "Trivy is set up as a GitHub Action and runs against each PR". I need to rework this action.

OK, if we _are_ scanning on PRs, and an image isn't built yet, then we need to build one here.

> Additionally, I'm not sure how useful it is to scan existing images: this will not help us be proactive about upcoming issues, and will only tell us existing issues which might already be fixed.

I actually think the prebuilt case is more interesting, as it detects newly discovered issues in images that we have already published, and are running in production. But my basic problem is scanning an image that we build on the fly. Also, we have no way to assert that what we built is the same as the image that's used when running our service.

.github/workflows/trivy.yml
fboulnois: Success here: https://github.com/DataBiosphere/terra-data-catalog/actions/runs/3002284454

pshapiro4broad (approved):

s-rubenstein (approved):
LGTM, although agreed with @pshapiro4broad

PR #123, opened 2022-09-06
DC-527: Enable Trivy scan on PR and nightly tests

fboulnois:
Requires https://github.com/DataBiosphere/terra-data-catalog/pull/122 to be merged.

PR action success: https://github.com/DataBiosphere/terra-data-catalog/actions/runs/3003505031
Trivy success: https://github.com/DataBiosphere/terra-data-catalog/actions/runs/3003512905

### DIFF
--- a/.github/workflows/trivy.yml
+++ b/.github/workflows/trivy.yml
@@ -45,7 +45,7 @@ jobs:
         with:
           channel: '#jade-data-explorer'
           status: failure
-          author_name: Build on dev
+          author_name: Trivy action
           fields: workflow,message
           text: 'Trivy scan failure :sadpanda:'
           username: 'Data Explorer GitHub Action'
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -62,18 +62,19 @@ jobs:
             --image=${{ steps.image-name.outputs.name }} \
             -Djib.console=plain
 
-#  dispatch-trivy:
-#    needs: [ build ]
-#    runs-on: ubuntu-latest
-#
-#    if: github.event_name == 'pull_request'
-#
-#    steps:
-#      - name: Fire off Trivy action
-#        uses: broadinstitute/workflow-dispatch@v1
-#        with:
-#          workflow: Trivy
-#          token: ${{ secrets.BROADBOT_TOKEN }}
+  dispatch-trivy:
+    needs: [ build ]
+    runs-on: ubuntu-latest
+
+    if: github.event_name == 'pull_request'
+
+    steps:
+      - name: Fire off Trivy action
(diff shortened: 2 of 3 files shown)


.github/workflows/trivy.yml
```diff
@@ -0,0 +1,53 @@
...
+        with:
+          java-version: '17'
+          distribution: 'temurin'
+          cache: 'gradle'
+
+      - name: Build all projects without running tests
+        run: ./gradlew --build-cache build -x test -x spotlessCheck
+
+      - name: Construct docker image name and tag
+        id: image-name
```
pshapiro4broad, line 21: Now that this is a completely separate operation from the PR workflow, I think it would be better to use a fixed name here, so it's clear that when an error is found, it's in the locally built image, and not in the image in artifcatory / gcr.

.github/workflows/trivy.yml
```diff
@@ -0,0 +1,53 @@
...
+    if: failure()
+    steps:
+      - name: Notify slack on failure
+        uses: broadinstitute/action-slack@v3.8.0
+        env:
+          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+        with:
+          channel: '#jade-data-explorer'
+          status: failure
+          author_name: Build on dev
```
pshapiro4broad, line 50: Is this the correct `author_name` for this workflow? It can be run on any branch.

.github/workflows/trivy.yml
```diff
@@ -0,0 +1,53 @@
...
+          fields: workflow,message
```
pshapiro4broad, line 51: Can you try using `job` instead of `workflow`? The intent here is we want a link to the failed workflow in GitHub; I think `job` will do this.

pshapiro4broad (approved):

.github/workflows/trivy.yml
fboulnois: If it fails, it links directly to the line that failed in the run, is this better than linking to the job? https://github.com/DataBiosphere/terra-data-catalog/commit/c4a369f473135f770183dbe21576bbe51723716e/checks

.github/workflows/trivy.yml
fboulnois: What do you suggest? Perhaps just `Trivy failure`? Or do we want to construct the environment as part of the message too?

.github/workflows/trivy.yml
fboulnois: I added a fixed name.

.github/workflows/trivy.yml
fboulnois: I named it Trivy failure for now.

s-rubenstein (approved):
LGTM

PR #124, opened 2022-09-06
DC-523: Fix missing content-length issue by adding a body to the POST

pshapiro4broad:
The swagger generated Java client library doesn't include a `Content-Length` header if the POST body is `null` (instead of providing `Content-Length: 0`, which our OIDC proxy server expects). There are ways to fix this by changing the client code behavior, and it may even be a bug in the swagger codegen, but instead of trying to fix the problem there, I think it's better to move the destination workflow ID parameter to a body parameter instead of a path parameter. We believe that the catalog will need to support different export destinations in the future, at which point we'd need to add more flexibility to the API anyway.

Since this endpoint is live in production, this change adds a new API and deprecates the existing one. Pushing this change out will require two follow-on PRs. One in the UI to use the new API, and the other is in the service to remove the deprecated endpoint.

### DIFF
--- a/integration/src/main/java/scripts/client/RawlsClient.java
+++ b/integration/src/main/java/scripts/client/RawlsClient.java
@@ -130,11 +130,13 @@ public class RawlsClient {
             .attributes(Map.of());
     var workspaceDetails = workspacesApi.createWorkspace(request);
     log.info("created workspace {}", workspaceDetails.getWorkspaceId());
-    ingestData(workspaceDetails.getNamespace(), workspaceDetails.getName());
     return workspaceDetails;
   }
 
-  private void ingestData(String namespace, String name) throws ApiException {
+  public void ingestData(WorkspaceDetails workspaceDetails) throws ApiException {
+    var namespace = workspaceDetails.getNamespace();
+    var name = workspaceDetails.getName();
+
     for (int i = 1; i <= 15; i++) {
       Entity entity = new Entity().entityType("sample").name("sample" + i);
       entity.setAttributes(
--- a/integration/src/main/java/scripts/testscripts/DatasetOperations.java
+++ b/integration/src/main/java/scripts/testscripts/DatasetOperations.java
@@ -15,6 +15,7 @@ import bio.terra.catalog.api.DatasetsApi;
 import bio.terra.catalog.client.ApiException;
 import bio.terra.catalog.model.ColumnModel;
 import bio.terra.catalog.model.CreateDatasetRequest;
+import bio.terra.catalog.model.DatasetExportRequest;
 import bio.terra.catalog.model.StorageSystem;
 import bio.terra.catalog.model.TableMetadata;
 import bio.terra.datarepo.model.DatasetModel;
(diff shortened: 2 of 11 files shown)


service/src/main/java/bio/terra/catalog/controller/DatasetApiController.java
```diff
@@ -20,22 +21,22 @@
 
 @Controller
 public class DatasetApiController implements DatasetsApi {
-  private final HttpServletRequest request;
+  private final HttpServletRequest servletRequest;
```
pshapiro4broad, line 24: This is an unrelated change to fix a warning in IntelliJ, where a method parameter with the same name (`request`) shadows the field.

integration/src/main/java/scripts/testscripts/DatasetOperations.java
```diff
@@ -67,6 +67,7 @@ public void setup(List<TestUserSpecification> testUsers) throws Exception {
   private void setupWorkspace(TestUserSpecification user) throws Exception {
     rawlsClient = new RawlsClient(server, user);
     workspaceSource = rawlsClient.createTestWorkspace();
+    rawlsClient.ingestData(workspaceSource);
```
pshapiro4broad, line 70: This change fixes a bug in the test code. Before it was ingesting data in both the source and destination workspaces, and the copy would fail since the destination workspace already had the same data as the source.

fboulnois (approved):
The OpenAPI and test changes lgtm.

integration/src/main/java/scripts/client/RawlsClient.java
```diff
@@ -130,11 +130,13 @@ public WorkspaceDetails createTestWorkspace() throws ApiException {
             .attributes(Map.of());
     var workspaceDetails = workspacesApi.createWorkspace(request);
     log.info("created workspace {}", workspaceDetails.getWorkspaceId());
-    ingestData(workspaceDetails.getNamespace(), workspaceDetails.getName());
     return workspaceDetails;
   }
 
-  private void ingestData(String namespace, String name) throws ApiException {
+  public void ingestData(WorkspaceDetails workspaceDetails) throws ApiException {
```
Abulu123, line 136: Looks good

Abulu123 (approved):
Pr looks good!

integration/src/main/java/scripts/testscripts/DatasetOperations.java
```diff
@@ -112,7 +111,7 @@ private void exportUserJourney(
 
     // Export workspace dataset to workspace
     var workspaceId = UUID.fromString(workspaceDest.getWorkspaceId());
-    datasetsApi.exportDataset(datasetId, workspaceId);
+    datasetsApi.exportDataset(datasetId, new DatasetExportRequest().workspaceId(workspaceId));
```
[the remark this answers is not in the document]
pshapiro4broad: My usual preference is to avoid local variables unless I think they make the code clearer. In general I'm not a fan of code that only does one thing per line and creates a new variable for each indeterminate step. On the other hand, writing code as a single line of text can obscure meaning and make the code hard to follow. So it's hard to say exactly when I think one is better or the other is, and the answer is probably different for everyone...

In this case I don't think it's necessary to introduce another variable. it would be great if we had better constructors for objects like `DataExportRequest`; I think the fluent style doesn't add much to code like this. But this is a swagger generated object so we're stuck with this API (or we need to do a lot of templating work to fix it).

PR #125, opened 2022-09-11
DC-512: Refactor dataset service code around a general storage system API

pshapiro4broad:
This change creates a general API for storage systems to implement to add their support to the catalog. All the TDR and Rawls specific code has been moved out of `DatasetService` and into `DatarepoService` or `RawlsService`.

The result is that `DatasetService` is easier to maintain and it's clearer where storage system specific code should go. This also adds a scaffold for external dataset support which can be extended when that feature is added.

There's some code duplication in the test code that should be refactored out before merging.


service/src/main/java/bio/terra/catalog/service/DatasetService.java
```diff
@@ -43,27 +36,35 @@ public class DatasetService {
...
     this.datasetDao = datasetDao;
     this.objectMapper = objectMapper;
   }
 
-  public static class IllegalMetadataException extends RuntimeException {
-    public IllegalMetadataException(Throwable cause) {
-      super(cause);
-    }
+  private StorageSystemService getService(StorageSystem system) {
+    return switch (system) {
```
pshapiro4broad, line 59: I'd like to get this out of this class but didn't see an easy way to do it, since enums instances are static and don't directly interact with spring injected objects. Maybe something for a future refactor.

service/src/main/java/bio/terra/catalog/service/DatasetService.java
```diff
@@ -131,22 +128,21 @@ private List<DatasetResponse> createDatasetResponses(
...
-
-  private List<DatasetResponse> collectWorkspaceDatasets(AuthenticatedUserRequest user) {
-    var roleMap = rawlsService.getWorkspaceInformation(user);
-    return createDatasetResponses(roleMap, StorageSystem.TERRA_WORKSPACE);
+    var roleMap = getService(system).getObjects(user);
+    return createDatasetResponses(roleMap, system);
   }
 
+  // FIXME: possibly refactor to first get all IDs, and then get all datasets. That will help
+  // FIXME: limit the DB work done here.
```
pshapiro4broad, line 140: I will probably make tech debt tickets for these FIXMEs and remove the comments, rather than fixing the issues now.

service/src/main/java/bio/terra/catalog/common/StorageSystemService.java
```diff
@@ -0,0 +1,72 @@
...
+  SystemStatusSystems status();
+
+  /**
+   * Return all objects in the storage system that are visible to the user, and the storage system
+   * information for each object.
+   *
+   * @param user the user
+   * @return a map of storage object ID to access level
+   */
+  Map<String, StorageSystemInformation> getObjects(AuthenticatedUserRequest user);
```
s-rubenstein, line 29: I _think_ there's a more descriptive name here than `getObjects`. Maybe `getStorageSystemData`?

s-rubenstein (approved):
This all looks reasonable to me. I like the helper methods in the service classes. I think I would want to make sure we addressed the FIXME's quickly, because they seem pretty useful to fix, and I think letting them linger might make them stale.

Is `FIXME` our preferred language? Or is `TODO`? (I have seen TODO mostly in the past, but don't think I've heard anything explicitly for this team.)

service/src/main/java/bio/terra/catalog/common/StorageSystemService.java
pshapiro4broad: I'd rather not repeat words that are already in the class name. So I don't think it should contain `storage` or `system`. Would `getData()` be better?  The other suggestion I had was `getDatasets()`, since it's returning the objects from the storage system that are datasets to the catalog.

service/src/main/java/bio/terra/catalog/service/DatasetService.java
pshapiro4broad: Done

petesantos (approved):

PR #126, opened 2022-09-13
DC-374: Update Gradle swagger plugin

fboulnois:
This PR migrates the swagger code generation from the `org.hidetake.swagger.generator` gradle plugin to the `org.openapitools` plugin.

Done
- unit and integration tests pass
- ~needed to change the metadata element in some web APIs from `string` to `object`; may want to change to `object` everywhere to be consistent~ reverting the `native` client change for now; it ended up causing too many other changes

Not Done
- findbugs/spotbugs scan is currently disabled
- ~some unit tests are still commented out~


service/src/test/java/bio/terra/catalog/datarepo/DatarepoServiceTest.java
```diff
@@ -101,23 +101,23 @@ void userHasActionException() throws Exception {
...
+    assertThat(datarepoService.status(), is(new SystemStatusSystemsValue().ok(true)));
   }
 
   @Test
   void statusDown() throws Exception {
     mockStatus();
     when(unauthenticatedApi.serviceStatus()).thenReturn(new RepositoryStatusModel().ok(false));
     var status = datarepoService.status();
-    assertFalse(status.isOk());
+    assertFalse(status.getOk());
```
pshapiro4broad, line 112: Could maybe avoid churn here by using `booleanGetterPrefix`, see https://openapi-generator.tech/docs/generators/java/.

integration/src/main/java/scripts/testscripts/DatasetOperations.java
```diff
@@ -173,43 +173,46 @@ private void assertDatasetValues(
...
+  private void crudUserJourney(StorageSystem storageSystem, String sourceId) throws Exception {
     // Given a snapshot, create a catalog entry.
     var request =
         new CreateDatasetRequest()
             .catalogEntry(createMetadata("crud").toString())
             .storageSourceId(sourceId)
             .storageSystem(storageSystem);
-    datasetId = datasetsApi.createDataset(request).getId();
-    assertThat(client.getStatusCode(), is(HttpStatusCodes.STATUS_CODE_OK));
+    var createResponse = datasetsApi.createDatasetWithHttpInfo(request);
```
s-rubenstein, line 183: How come this is `.createDatasetWithHttpInfo` rather than `createDataset`? Did the operation id change? Or is this a new method that returns the full http object instead of just the response body?

integration/src/main/java/scripts/testscripts/DatasetPermissionOperations.java
```diff
@@ -121,14 +121,12 @@ private void testDatasetReaderPermissions(
       throws ApiException {
     CreateDatasetRequest request = datasetRequest(storageSourceId, storageSystem);
     // note if this assertion fails we'll leave behind a stray dataset in the db
-    assertThrows(ApiException.class, () -> userDatasetsApi.createDataset(request));
-    assertThat(
-        userDatasetsApi.getApiClient().getStatusCode(),
-        is(HttpStatusCodes.STATUS_CODE_UNAUTHORIZED));
+    var exception = assertThrows(ApiException.class, () -> userDatasetsApi.createDataset(request));
```
s-rubenstein, line 124: Why isn't this using the `createDatasetWithHttpInfo` that was used above? Is it because the response body includes the status code?

s-rubenstein (commented):
This all looks reasonable to me!

integration/src/main/java/scripts/testscripts/DatasetOperations.java
pshapiro4broad: Yes, although the library is the same (`jersey2`) the client codegen is still different. One difference is that the HTTP code isn't present in `client` anymore. So instead we need to use the `.xxxWithHttpInfo()` API variants.

integration/src/main/java/scripts/testscripts/DatasetPermissionOperations.java
pshapiro4broad: It's because this throws an exception. A non-20x response code will cause an `ApiException` to be thrown, so it needs to be called inside `assertThrows()` or some other `try` form; even if we called the `WithHttpInfo()` version here it would still throw and never return a response. Once we have the exception we can look at the http code in the exception (which I think we would've done before as well).

pshapiro4broad:
This turned out to be a lot more work than I had hoped. I created another ticket to track the work of porting our client and server codegen to newer versions; once that's done, it should be easier to migrate to the different plugin.

PR #127, opened 2022-09-16
Use 403 instead of 401 now that TDR is updated

fboulnois:


s-rubenstein (approved):
LGTM

integration/src/main/java/scripts/testscripts/DatasetPermissionOperations.java
```diff
@@ -165,8 +165,7 @@ private void testDiscovererPermissionsForSnapshotDataset() throws Exception {
         ApiException.class,
         () -> userDatasetsApi.listDatasetPreviewTables(adminTestSnapshotDatasetId));
     assertThat(
-        userDatasetsApi.getApiClient().getStatusCode(),
-        is(HttpStatusCodes.STATUS_CODE_UNAUTHORIZED));
+        userDatasetsApi.getApiClient().getStatusCode(), is(HttpStatusCodes.STATUS_CODE_FORBIDDEN));
```
pshapiro4broad, line 168: Should this test explicitly require 403, or should our test use `is4xxClientError()` instead?

Is the 403 response part of our API that we must preserve (and have full control over), or is it a way for us to expose the behavior of another service to the caller?

If it's part of another service's behavior (which I think it is), then I think our test shouldn't rely on a specific 4xx error code.

We may want to test that 4xx codes are properly propagated by our APIs, but that can be done in a unit test where both the input and output codes are controlled by the test.

pshapiro4broad (approved):
This change is fine, and fixes the immediate issue, but I wonder if we should make our test less tied to another service's behavior, since that behavior is not depended on by catalog itself. Making it less dependent would make the test less fragile; right now we have a state where the service itself is fine, and the code is fine, but our tests are failing.

PR #128, opened 2022-09-16
[DC-529] phs request access url

s-rubenstein:


fboulnois (approved):
lgtm, reviewed in mobbing

integration/src/main/java/scripts/testscripts/DatasetOperations.java
```diff
@@ -161,12 +164,26 @@ private void previewUserJourney(StorageSystem storageSystem, String sourceId)
     datasetId = null;
   }
 
+  private ObjectNode expectedMetadataResponse(ObjectNode metadata, StorageSystem storageSystem) {
+    ObjectNode expectedMetadata = metadata.deepCopy().put("id", datasetId.toString());
```
pshapiro4broad, line 168: Rather than using `deepCopy()` I think it'd be better to move the metadata out of a static and into a regular field. Having mutable objects like `ObjectNode` as static members isn't great in general IMO and we don't want test writers to have to think about when they might need to copy them when writing a test.

integration/src/main/java/scripts/testscripts/DatasetOperations.java
```diff
@@ -161,12 +164,26 @@ private void previewUserJourney(StorageSystem storageSystem, String sourceId)
...
+    if (storageSystem.equals(StorageSystem.TDR)) {
+      expectedMetadata
+          .put("phsId", "1234")
+          .put(
+              "dcat:accessURL",
+              String.format(
+                  "https://www.ncbi.nlm.nih.gov/projects/gap/cgi-bin/study.cgi?study_id=%s",
```
pshapiro4broad, line 175: I don't think we should hardcode this path in the test, as the path contents isn't part of the API; the API specifies that the property must be present but doesn't specify what its contents are. Having the test rely on the contents means that the test must be updated if the contents ever change. There are a few ways to match json objects, we could check for each property separately or try using one of the json matching libraries, like https://github.com/skyscreamer/JSONassert.

integration/src/main/java/scripts/testscripts/DatasetOperations.java
```diff
@@ -161,12 +164,26 @@ private void previewUserJourney(StorageSystem storageSystem, String sourceId)
     datasetId = null;
   }
 
+  private ObjectNode expectedMetadataResponse(ObjectNode metadata, StorageSystem storageSystem) {
+    ObjectNode expectedMetadata = metadata.deepCopy().put("id", datasetId.toString());
+    if (storageSystem.equals(StorageSystem.TDR)) {
+      expectedMetadata
+          .put("phsId", "1234")
+          .put(
+              "dcat:accessURL",
```
pshapiro4broad, line 173: I don't think this is the correct property for this attribute. This isn't the "access URL", it's the "request access URL". https://www.w3.org/TR/vocab-dcat-3/#Property:distribution_access_url.

Using the same property as accessURL would also mean that we couldn't use it to determine if the URL was supposed to be used for Request Access or not. Additionally, I think it's plausible that a dataset will have both a request access URL and an access URL, and that they'd be different.

integration/src/main/java/scripts/testscripts/DatasetOperations.java
```diff
@@ -161,12 +164,26 @@ private void previewUserJourney(StorageSystem storageSystem, String sourceId)
     datasetId = null;
   }
 
+  private ObjectNode expectedMetadataResponse(ObjectNode metadata, StorageSystem storageSystem) {
+    ObjectNode expectedMetadata = metadata.deepCopy().put("id", datasetId.toString());
+    if (storageSystem.equals(StorageSystem.TDR)) {
+      expectedMetadata
+          .put("phsId", "1234")
```
pshapiro4broad, line 171: I don't think `phsId` should be exposed as a property, unless there's a specific need in the UI for this attribute.

pshapiro4broad: It sounds like there definitely is a UI need!

service/src/test/java/bio/terra/catalog/service/DatasetServiceTest.java
```diff
@@ -173,6 +201,16 @@ void testGetMetadata() {
     verify(datasetDao).retrieve(datasetId);
   }
 
+  @Test
+  void testGetMetadataWithPhsId() {
+    when(datasetDao.retrieve(tdrDataset.id())).thenReturn(tdrDataset);
+    when(samService.hasGlobalAction(user, SamAction.READ_ANY_METADATA)).thenReturn(true);
+    when(datarepoService.getSnapshotPhsId(user, tdrDataset.storageSourceId())).thenReturn("1234");
+    String datasetMetadata = datasetService.getMetadata(user, tdrDataset.id());
+    verify(datasetDao).retrieve(tdrDataset.id());
```
pshapiro4broad, line 210: Is this `verify()` needed? The test should fail if no mocks are called, and since this test has an `assert`, sonar won't complain about the test not having any asserts.

service/src/test/java/bio/terra/catalog/datarepo/DatarepoServiceTest.java
```diff
@@ -54,12 +56,32 @@ private void mockStatus() {
   @Test
   void getSnapshots() throws Exception {
     mockSnapshots();
-    var items = Map.of("id", List.of("steward"));
-    var expectedItems = Map.of("id", DatasetAccessLevel.OWNER);
-    var esm = new EnumerateSnapshotModel().roleMap(items);
+    UUID snapshotId = UUID.randomUUID();
```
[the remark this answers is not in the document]
s-rubenstein: On the line below `.items(List.of(new SnapshotSummaryModel().id(snapshotId).phsId("1234")))`, snapshotId is required to be a UUID

service/src/test/java/bio/terra/catalog/service/DatasetServiceTest.java
```diff
@@ -123,10 +134,27 @@ void listDatasets() {
     assertThat(tdrJson.get("accessLevel").asText(), is(String.valueOf(DatasetAccessLevel.OWNER)));
   }
 
+  @Test
+  void listDatasetsWithPhsId() {
```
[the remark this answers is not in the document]
s-rubenstein: I think personally that it is useful to test both cases. This makes it more separated and will make it easier to debug (in my opinion) when something is broken. It also means that we have a test failure still if we start populating phsId when it doesn't exist.

service/src/test/java/bio/terra/catalog/service/DatasetServiceTest.java
```diff
@@ -173,6 +201,16 @@ void testGetMetadata() {
     verify(datasetDao).retrieve(datasetId);
   }
 
+  @Test
+  void testGetMetadataWithPhsId() {
```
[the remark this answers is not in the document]
s-rubenstein: (Same as above response - I think that it is useful to keep our tests more self contained)

pshapiro4broad (approved):

PR #130, opened 2022-09-22
Oa dc 464 schema validation

Abulu123:
Methods working. Moving onto testing.

### DIFF
--- a/common/openapi.yml
+++ b/common/openapi.yml
@@ -363,9 +363,6 @@ components:
       properties:
         name:
           type: string
-        array_of:
-          type: boolean
-          description: if true, this column is an array of data.
 
     TableMetadata:
       required:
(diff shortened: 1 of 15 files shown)


common/openapi.yml
```diff
@@ -363,9 +363,6 @@ components:
       properties:
         name:
           type: string
-        array_of:
```
pshapiro4broad, line 366: It looks like this branch is out of date, as this change is already on `main`. Maybe try to rebase your changes on `main` to fix this? Once it's fixed you shouldn't have changes in this file, `DatasetOperations.java`, `RawlsClient.java` and a few more.

pshapiro4broad: My guess is this happened because you started work on an existing branch, rather than first checking out `main` and creating a new branch from there. This happens sometimes by accident, I usually fix it by cherry-picking my changes onto a new branch off of `main`.

PR #131, opened 2022-09-22
Oa dc 464 schema validation2

Abulu123:


service/src/main/java/bio/terra/catalog/service/SchemaService.java
```diff
@@ -0,0 +1,38 @@
...
+import com.networknt.schema.SpecVersion;
+import com.networknt.schema.ValidationMessage;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.util.Set;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+
+@Service
+public class SchemaService {
```
pshapiro4broad, line 17: This might not be the best name.. this is a service that validates dataset metadata against a fixed schema. So it's not really a general schema service but one specific to dataset metadata. Although I guess in the future we might have other schemas that we validate with? So maybe `JsonValidationService`, since it what it does is validate json. The fact that it uses a schema to do it isn't necessarily important to the caller.

service/src/main/java/bio/terra/catalog/service/SchemaService.java
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  private JsonSchema getJsonSchemaFromUrl(String uri) throws FileNotFoundException {
+    JsonSchemaFactory factory = JsonSchemaFactory.getInstance(SpecVersion.VersionFlag.V4);
+    return factory.getSchema(new FileInputStream(uri));
+  }
+
+  public void validateMetadata(JsonNode json) {
+    Set<ValidationMessage> errors = schema.validate(json);
+    if (errors.size() > 0) {
```
pshapiro4broad, line 34: In general I think it's better to use `!isEmpty()` instead of `> 0`. The two have the same effect (as would `!= 0`) but `isEmpty()` is a higher level concept.
```suggestion
    if (!errors.isEmpty()) {
```

service/src/main/resources/schema.json
```diff
@@ -0,0 +1,278 @@
...
+    "dcat:accessURL": {
+      "type": "string",
+      "description": "A URL to the dataset in the underlying storage system"
+    },
+    "TerraDCAT_ap:hasDataUsePermission": {
+      "description": "Full List taken from https://www.ebi.ac.uk/ols/ontologies/duo/terms?iri=http://purl.obolibrary.org/obo/DUO_0000001",
+      "enum": [
+        "TerraCore:NoRestriction",
+        "TerraCore:GeneralResearchUse",
+        "TerraCore:NPOA",
```
pshapiro4broad, line 50: Are these fields correct? I thought we found that they aren't the correct values to match on for this property. @tcjiang9

service/src/main/resources/application.yml
```diff
@@ -74,6 +74,9 @@ catalog:
   rawls:
     basePath: ${env.rawls.basePath}
 
+  schema:
+    basePath: src/main/resources/schemas/catalog-v0.schema.json
```
pshapiro4broad, line 78: I don't see the commit for this path in the PR

service/src/test/java/bio/terra/catalog/service/DatasetServiceTest.java
```diff
@@ -96,9 +99,10 @@ class DatasetServiceTest {
           null);
 
   @BeforeEach
-  public void beforeEach() {
+  public void beforeEach() throws URISyntaxException {
```
pshapiro4broad, line 102: When writing tests, I think it's better to use `throws Exception` since we don't want to be updating our tests whenever a checked exception in an API is changed.

service/src/main/java/bio/terra/catalog/service/SchemaService.java
```diff
@@ -0,0 +1,38 @@
...
+import org.springframework.stereotype.Service;
+
+@Service
+public class SchemaService {
+
+  private final JsonSchema schema;
+
+
+  @Autowired
+  public SchemaService(SchemaConfiguration schemaConfiguration) throws FileNotFoundException {
```
pshapiro4broad, line 23: I don't think an injected object should throw a checked exception. I would catch and re-throw this as a custom unchecked exception, such as `SchemaServiceInitializationException`.

service/src/main/java/bio/terra/catalog/service/SchemaService.java
```diff
@@ -0,0 +1,38 @@
...
+    this.schema = getJsonSchemaFromUrl(schemaConfiguration.basePath());
+  }
+
+  private JsonSchema getJsonSchemaFromUrl(String uri) throws FileNotFoundException {
+    JsonSchemaFactory factory = JsonSchemaFactory.getInstance(SpecVersion.VersionFlag.V4);
+    return factory.getSchema(new FileInputStream(uri));
```
pshapiro4broad, line 29: This can (will?) leave an input stream open. Instead you should use a `try` block to make sure that all opened streams are closed, even if an exception occurs:
```suggestion
    try (var input = new FileInputStream(uri)) {
      return factory.getSchema(input);
    }
```

service/src/main/java/bio/terra/catalog/service/SchemaService.java
```diff
@@ -0,0 +1,38 @@
...
+  private final JsonSchema schema;
+
+
+  @Autowired
+  public SchemaService(SchemaConfiguration schemaConfiguration) throws FileNotFoundException {
+    this.schema = getJsonSchemaFromUrl(schemaConfiguration.basePath());
+  }
+
+  private JsonSchema getJsonSchemaFromUrl(String uri) throws FileNotFoundException {
+    JsonSchemaFactory factory = JsonSchemaFactory.getInstance(SpecVersion.VersionFlag.V4);
```
pshapiro4broad, line 28: Is there a reason why this specific version is used over others? It looks like `SpecVersion.VersionFlag.V201909` is the most recent spec version supported.

service/src/main/java/bio/terra/catalog/service/SchemaService.java
```diff
@@ -0,0 +1,38 @@
...
+
+  private final JsonSchema schema;
+
+
+  @Autowired
+  public SchemaService(SchemaConfiguration schemaConfiguration) throws FileNotFoundException {
+    this.schema = getJsonSchemaFromUrl(schemaConfiguration.basePath());
+  }
+
+  private JsonSchema getJsonSchemaFromUrl(String uri) throws FileNotFoundException {
```
pshapiro4broad, line 27: The code calls this both `uri` and `url` but it looks like it's a file name, not a `uri` or `url`.

service/src/main/resources/schemas/test.schema.json
```diff
@@ -0,0 +1,15 @@
+{
```
pshapiro4broad, line 1: This shouldn't be in the `src/main/resources` directory, it should be in the `src/test/resources` directory, since it's a resource only used by tests and is not part of the deployed service.

pshapiro4broad (commented):
A few more comments. Also I think I was wrong about integration tests, in that I think we need them, to ensure that the application configuration is set up correctly.

Actually I'm wondering if the integrations should be failing right now with these changes, since I don't think they provide all of the required fields.

.github/workflows/build-and-test.yml
```diff
@@ -182,7 +182,7 @@ jobs:
           fi
 
       - name: Run the integration test suite
-        run: ./gradlew --build-cache runTest --args="suites/dev/FullIntegration.json build/reports"
+        run: ./gradlew --build-cache runTest --args="suites/local/FullIntegration.json build/reports"
```
pshapiro4broad, line 185: It looks like your branch is out of date with `main` in some way. This diff shouldn't be showing up in your PR.

scripts/build.gradle
```diff
@@ -14,7 +14,7 @@ shellcheck {
 spotless {
     python {
         target '**/*.py'
-        black()
+        black('22.6.0')
```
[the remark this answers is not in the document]
pshapiro4broad: You still need to revert this change

service/src/main/java/bio/terra/catalog/service/SchemaService.java
pshapiro4broad, line 17: It looks like when you renamed the class, you somehow lost the file rename, so the original file didn't get deleted. You need to add another commit that removes this file.

service/src/main/resources/schema.json
```diff
@@ -0,0 +1,278 @@
+{
```
pshapiro4broad, line 1: This still needs to be deleted too

service/src/main/resources/schemas/test.schema.json
```diff
@@ -0,0 +1,15 @@
...
+  "title": "Catalog Schema V1",
```
pshapiro4broad, line 2: I would use different text here to make it clear that this is a test schema

service/src/test/java/bio/terra/catalog/service/DatasetServiceTest.java
pshapiro4broad: This doesn't seem to be fixed yet.

service/src/test/java/bio/terra/catalog/service/JsonValidationServiceTest.java
```diff
@@ -0,0 +1,32 @@
...
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import org.junit.jupiter.api.Test;
+
+public class JsonValidationServiceTest {
+  private static final String BASE_PATH = "src/test/resources/schemas/test.schema.json";
+
+  private static final ObjectMapper objectMapper = new BeanConfig().objectMapper();
+
+  @Test
+  void testValidateInvalidMetadata() {
```
pshapiro4broad, line 19: Can you add another test that verifies that an invalid schema will throw an error during object construction? Once you've tested that, you should have 100% coverage of the new service code. You can look at the coverage of the test by running it in IntelliJ.

service/src/test/java/bio/terra/catalog/service/SchemaServiceTest.java
```diff
@@ -0,0 +1,33 @@
...
+import bio.terra.catalog.config.BeanConfig;
+import bio.terra.catalog.config.SchemaConfiguration;
+import bio.terra.common.exception.BadRequestException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import java.io.FileNotFoundException;
+import org.junit.jupiter.api.Test;
+
+public class SchemaServiceTest {
```
pshapiro4broad, line 14: This also needs to be deleted.

integration/src/main/java/scripts/testscripts/DatasetOperations.java
```diff
@@ -83,8 +84,32 @@ private void setupSnapshot(TestUserSpecification user) throws Exception {
...
+      String date,
+      List DataCollections,
+      String url,
+      List generatedBy,
+      List storage,
+      Object counts,
+      List contributors) {
+    var obj =
+        objectMapper
+            .createObjectNode()
```
pshapiro4broad, line 100: At this point I think it might be easier to create a `Map` here and then use `objectMapper` to convert the `Map` to `String`. Using `Map` would let you create most of the object's content using `Map.of()`.

integration/src/main/java/scripts/testscripts/DatasetOperations.java
```diff
@@ -107,7 +132,19 @@ private void exportUserJourney(
...
     var request =
         new CreateDatasetRequest()
-            .catalogEntry(createMetadata("export-test-dataset").toString())
+            .catalogEntry(
+                createMetadata(
+                        "crud",
+                        "description",
+                        "creator",
+                        "date",
+                        new ArrayList(),
```
pshapiro4broad, line 141: Unless you need to create a modifiable list, you should use `List.of()` instead of `new ArrayList<>()` to create an empty list.

service/src/main/resources/schema/development/schema.json
```diff
@@ -21,11 +21,11 @@
       "type": "string"
     },
     "dct:issued": {
-      "type": "date-time",
```
pshapiro4broad, line 24: Did this change because the schema version changed, so it's not supported? Or was `date-time` not correct to begin with? @tcjiang9

integration/src/main/java/scripts/testscripts/DatasetOperations.java
```diff
@@ -107,7 +121,7 @@ private void exportUserJourney(
     // Create workspace dataset
     var request =
         new CreateDatasetRequest()
-            .catalogEntry(createMetadata("export-test-dataset").toString())
+            .catalogEntry(createMetadata("crud").toString())
```
pshapiro4broad, line 124: I think you can revert this change

pshapiro4broad (commented):
One last thing I noticed, there's a change that I think is unnecessary and can be reverted.

pshapiro4broad (approved):

s-rubenstein (approved):
This looks great!

pshapiro4broad:
Before you merge, please fix the "code smell" reported by sonar, it looks like an easy fix.

PR #132, opened 2022-09-29
DC-353 v1 schema

tcjiang9:
still need to improve some descriptions


s-rubenstein (commented):
I think we probably want the path here to be in resources? And also, I would put it in a schemas folder, and call it `v1.json`, rather than `schema.json`?

Other than that, it looks good! Although per standup conversation, did we want to add a string field for Brand to this?

service/src/main/java/bio/terra/catalog/schema.json
```diff
@@ -0,0 +1,192 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
```
pshapiro4broad, line 2: I think an `$id` field is required here.

service/src/main/java/bio/terra/catalog/schema.json
```diff
@@ -0,0 +1,192 @@
...
+  "title": "Catalog Schema V1",
```
pshapiro4broad, line 3: Does it make sense to put the version in the title? Looking at example schemas at json-schema.org, it looks like schemas have names that reflect what the represent. In this case, this represents a `catalog dataset`, or maybe just `dataset` if the schema `id` is scoped to the catalog service.

service/src/main/java/bio/terra/catalog/schema.json
```diff
@@ -0,0 +1,192 @@
...
+  "type": "object",
+  "required": ["dct:title", "dct:description", "dct:creator", "dct:issued", "TerraDCAT_ap:hasDataCollection", "dcat:accessURL", "prov:wasGeneratedBy", "storage", "counts", "contributors"],
+  "properties": {
```
pshapiro4broad, line 6: I don't see `requestAccessURL` in this list of properties, I think it should be there. It should also be in the v0 schema.

pshapiro4broad (commented):
I can't recall if we've talked about this before or not, but I think the schema should be in a directory or path with the version in its name. For example: `schema/v1/schema.json` or `schema/schema-v1.json`.

Also, the schema should be in the `resources` directory, not the `java` directory (`service/src/main/resources`).

I don't think the v0 schema is committed yet, if not, that should also be committed too

Finally, at any time I think we want a `development` schema which is basically the next released schema as it's being worked on. So if both v1 and v0 are checked in, we'd have a third for development work. That could be the same as v1 to begin with. Although maybe this should really be `development` until the radx ingest work is far enough along to know that we're properly handling the radx cases.

service/src/main/java/bio/terra/catalog/schema.json
s-rubenstein: I was wondering if we should put phsId in here as well, then, but I think we shouldn't because it's stored solely in TDR, and we don't want users adding it here. If we wanted to support phsId for workspaces that would change.
(there's also access level and id, but I think we don't need those because users aren't doing anything with them)

service/src/main/java/bio/terra/catalog/schema.json
```diff
@@ -86,6 +86,10 @@
     "contributors": {
       "type": "array",
       "items": { "type": "string" }
+    },
+    "brands": {
```
s-rubenstein, line 90: Talking to @pshapiro4broad, we may want to call this `consortiums` or something? Just thinking about it, a brand is a UI concept, but a consortium is actually what RADx/RAREx/etc is.

pshapiro4broad: Yes, I don't think `brand` is correct for dataset metadata since a brand may include data from a mix of different sources, and the dataset itself shouldn't need to know about which brands corresponds to which source. My view on things:

- A dataset has a value that describes the data source, this is "consortium" or "study" or whatever value makes sense. I think the existing "consortium" field should be used unless we can think of a reason why it won't work.
- Brand is a UI concept which may correspond to a set of filters on the data shown. A brand will need to know which consortiums (or studies) it can display.

pshapiro4broad: For the record, what the UI shows as `Consortium` is computed from the model by looking at the `dct:identifier` property of the first element of the array property `TerraDCAT_ap:hasDataCollection`. So my suggestion is to store the RADx study  in `TerraDCAT_ap:hasDataCollection`. (The code could use `dct:title` instead of `dct:identifier` if this makes more sense; right now no data in production is using this field.)

s-rubenstein (approved):
LGTM

pshapiro4broad (approved):

PR #135, opened 2022-10-05
DC-554 Follow-on fixes to integration tests

pshapiro4broad:
- don't notify `#dsde-qa` on `dev` test runs
- rename script to `integration` instead of `nightly`

An example of a test run that skips messaging `dsde-qa` is https://github.com/DataBiosphere/terra-data-catalog/actions/runs/3192183584


s-rubenstein (approved):
LGTM

PR #136, opened 2022-10-11
DC-542: fix regression where container main class was moved but config wasn't updated

pshapiro4broad:
Since our PR tests don't run the container, no error was generated before the merge.


s-rubenstein (approved):
LGTM

petesantos (approved):

PR #137, opened 2022-10-13
DC-592: GitHub action fixes: remove use of `set-output`

pshapiro4broad:
Also 
- update actions to newer versions to fix node version warning
- fix issue where failure to start server didn't result in boot logs being printed out
- increase test timeout


s-rubenstein (approved):
LGTM

tcjiang9 (approved):

PR #138, opened 2022-10-14
D-490 make schema edits after meeting with Nathan

tcjiang9:


service/src/main/resources/schema/development/schema.json
```diff
@@ -138,6 +138,17 @@
         }
       }
     },
+    "generatedBy": {
+      "type": "object",
+      "properties": {
+        "TerraCore:hasAssayCategory": {
```
pshapiro4broad, line 144: For these categories, do we have recommended dictionaries or ontologies? I think we want different catalog entries to agree on these terms, so either restricting the list, or providing a list of recommendations would be helpful.

service/src/main/resources/schema/development/schema.json
```diff
@@ -180,6 +191,19 @@
           "type": "integer"
         }
       }
+    },
+    "samples": {
+      "type": "object",
+      "properties": {
+        "disease": {
```
pshapiro4broad, line 198: I think we need dictionaries for these too. We don't need necessarily need them for this checkin but we'd want to start the process so we can have a list in the future.

service/src/main/resources/schema/development/schema.json
```diff
@@ -69,8 +68,8 @@
     },
     "prov:wasGeneratedBy": {
```
pshapiro4broad, line 69: What does it mean to have `prov:wasGeneratedBy` in `required` above if it's an array type? Can it be an empty array? Do we want instead to require at least one element (if that's possible)?

pshapiro4broad (commented):
I have some questions about the schema, see my comments.

service/src/main/resources/schema/development/schema.json
tcjiang9: Note, require minimum size

service/src/main/resources/schema/development/schema.json
pshapiro4broad: @kreinold says, we can remove this requirement / field for now, as we don't think it will be generally useful for searchability.

service/src/main/resources/schema/development/schema.json
```diff
@@ -138,6 +138,17 @@
...
       }
     },
+    "generatedBy": {
+      "type": "object",
+      "properties": {
+        "TerraCore:hasAssayCategory": {
+          "type": "string"
+        },
+        "TerraCore:hasDataModality": {
+          "type": "string"
```
pshapiro4broad, line 148: @rsc3 can you keep us up to date on the development of the data modality ontology?

The goal is to have a dictionary of terms we can use to add to our schema to help data ingesters conform to a common model.

service/src/main/resources/schema/development/schema.json
```diff
@@ -180,6 +191,19 @@
...
+    },
+    "samples": {
+      "type": "object",
+      "properties": {
+        "disease": {
+          "type": "array",
+          "items": { "type": "string" }
+        },
+        "species": {
+          "type": "array",
```
pshapiro4broad, line 203: For this, we should use https://www.ebi.ac.uk/ols/ontologies/ncit/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FNCIT_C14250&lang=en&viewMode=All&siblings=false

Maybe we should add a link to this to the description, not sure how best to add it to the json schema

service/src/main/resources/schema/development/schema.json
```diff
@@ -180,6 +191,19 @@
           "type": "integer"
         }
       }
+    },
+    "samples": {
+      "type": "object",
+      "properties": {
+        "disease": {
+          "type": "array",
```
pshapiro4broad, line 199: For this, we should use https://www.ebi.ac.uk/ols/ontologies/mondo/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FMONDO_0000001&lang=en&viewMode=PreferredRoots&siblings=false

pshapiro4broad (approved):
This looks OK, although I think we should start documenting the dictionary or ontology for fields when it's known.

In the case of ontologies, it may not make sense to copy it into our schema, since an ontology can change, and can be very large.

On the other hand, if the schema is supposed to capture what's known at a fixed point in time, maybe we could copy them? Something to consider, if there's no easy way to use the ontology directly inside a json schema.

This page may be very helpful too. It shows a list of field types and what ontology is used for each. It doesn't (yet) show which field has which type though. https://github.com/DataBiosphere/terra-interoperability-model/blob/master/documents/Ontologies.md

service/src/main/resources/schema/development/schema.json
```diff
@@ -152,7 +165,7 @@
           "enum": ["bigquery", "firestore", "bucket"]
         },
         "cloudPlatform": {
-          "enum": ["gcp"]
+          "enum": ["gcp", "azure"]
```
s-rubenstein, line 168: If we are adding azure, we probably need to add more enums to cloudResource and region? I don't know what those are, but we should check if we are adding the azure as a platform.

PR #140, opened 2022-10-21
DC-464: Fix issue where path to resource didn't use a resource path lookup

pshapiro4broad:
Using a fixed file path works in the source tree but doesn't work in the .jar deployment. Changing to use a resource reference using `classpath:` should fix this.

Also
- generate a date-time value on the fly using `Instant`
- add an example of date-time to the test schema
- refactor duplicate code in integration tests that creates dataset metadata

### DIFF
--- a/service/src/main/resources/application.yml
+++ b/service/src/main/resources/application.yml
@@ -75,7 +75,7 @@ catalog:
     basePath: ${env.rawls.basePath}
 
   schema:
-    basePath: src/main/resources/schema/development/schema.json
+    basePath: classpath:schema/development/schema.json
 
 terra.common:
   kubernetes:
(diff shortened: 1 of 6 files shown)


s-rubenstein (approved):
LGTM

service/src/main/resources/application.yml
```diff
@@ -75,7 +75,7 @@ catalog:
     basePath: ${env.rawls.basePath}
 
   schema:
-    basePath: src/main/resources/schema/development/schema.json
+    basePath: classpath:schema/development/schema.json
```
okotsopoulos, line 78: Not a blocker to merge, but I'm used to a `basePath` referring to a partial path common to many resources, like how y'all define a `basePath` for interacting with Rawls API above.  This might have a more appropriate name.

okotsopoulos (approved):

service/src/main/resources/application.yml
s-rubenstein: I'll have Phil look into that on Monday, when he's back, and move forwards with the PR as it stands to unblock monolith release.

pshapiro4broad: I agree, the name isn't quite right. I can PR a fix or @Abulu123 could fix it.

PR #145, opened 2022-11-03
DC-560: On PRs, run integration tests with just-built docker container

pshapiro4broad:
This changes our PR process to run integration tests using the jib-built docker image instead of starting a server using `gradle bootRun`. This has the benefit of finding problems with the docker build process before a PR is merged up.

As part of this work, the `jib` job is now a step inside the `integration-tests` job. It could be done separately, but no other job uses the built image, and uploading/downloading the image as an artifact takes an extra ~2 minutes.

A further improvement on this process would be to deploy the built image to a GKE cluster and run the tests there, possibly using a BEE. That would also help debug our helm deployment configuration, and provide a more realistic test environment.

To see an example of a failed run with the boot run log output: https://github.com/DataBiosphere/terra-data-catalog/actions/runs/3565847046/jobs/5991554899

Example of failed run due to incorrect jib image creation: https://github.com/DataBiosphere/terra-data-catalog/actions/runs/3567373980/jobs/5995087936

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -184,7 +184,7 @@ jobs:
       - name: Wait for boot run to be ready
         run: |
           set +e
-          timeout 60 bash -c 'until echo > /dev/tcp/localhost/8080; do sleep 1; done'
+          timeout 60 bash -c 'until [[ $(curl localhost:8080/status | jq .ok) == "true" ]]; do sleep 1; done'
           resultStatus=$?
           set -e
           if [[ $resultStatus == 0 ]]; then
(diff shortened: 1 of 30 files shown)


petesantos (approved):

s-rubenstein (approved):
Do we have a sense of how long this adds to the runtime of our integration tests? Is that the 2 minutes called out in the PR description?

pshapiro4broad:
@s-rubenstein 
> Do we have a sense of how long this adds to the runtime of our integration tests? Is that the 2 minutes called out in the PR description?

If only GitHub workflows was a real CI system then we could see some nice plots of job duration over time...

Looking at a bunch of passing test runs from this branch, integration tests take about 8 min 30 sec. Looking at builds from other PRs, the integration test job currently takes about 8 min 10 sec. So this appears to add **about 20s overhead** to the integration test and the build workflow overall.

It looks like the web app startup is faster, since it skips gradle, but now the `jib` build must be done before running the integration tests. Before it was done in parallel in a different job, so its time didn't affect the tests.

PR #146, opened 2022-11-03
DC-566: sonar coverage

Abulu123:
This Ticket involved me improving the unit test coverage for `DaoKeyHolder`, `SamService`, `BaseStatusService`,  and `StorageSystem`. `DaoKeyHolder` and `StorageSystem` didn't have unit tests so I had to create some. They are rather simple tests for specific cases in the code.


service/src/main/java/bio/terra/catalog/common/DaoKeyHolder.java
```diff
@@ -19,10 +19,7 @@ public Timestamp getTimestamp(String fieldName) {
 
   public Instant getCreatedDate() {
     Timestamp timestamp = getTimestamp("created_date");
-    if (timestamp != null) {
```
s-rubenstein, line 22: Why can we do this? Is `created_date` non nullable?

service/src/main/java/bio/terra/catalog/common/DaoKeyHolder.java
```diff
@@ -31,10 +28,9 @@ public String getString(String fieldName) {
...
   public <T> T getField(String fieldName, Class<T> type) {
     Map<String, Object> keys = getKeys();
-    if (keys != null) {
-      Object fieldObject = keys.get(fieldName);
-      return type.cast(fieldObject);
-    }
-    return null;
+    // In practice getKeys() will never return null
+    @SuppressWarnings("ConstantConditions")
+    Object fieldObject = keys.get(fieldName);
```
s-rubenstein, line 33: If we aren't null checking, could we inline `getKeys()`?

service/src/test/java/bio/terra/catalog/common/StorageSystemTest.java
```diff
@@ -0,0 +1,12 @@
+package bio.terra.catalog.common;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+import org.junit.jupiter.api.Test;
+
+class StorageSystemTest {
+  @Test
+  void fromStringIllegalArgumentException() {
```
s-rubenstein, line 9: Could we add a positive test here as well? (Something `fromString` that takes a valid value)

service/src/test/java/bio/terra/catalog/service/BaseStatusServiceTest.java
```diff
@@ -23,4 +26,23 @@ void getCurrentStatus() {
...
+    service.registerStatusCheck("", status);
+    service.startStatusChecking();
+    verify(status).get();
+  }
+
+  @Test
+  void getNonEnabledStatus() {
+    var config = new StatusCheckConfiguration(false, 0, 0, 10);
+    BaseStatusService service = new BaseStatusService(config);
+    assertThat(service.getCurrentStatus(), is(new SystemStatus().ok(true)));
```
s-rubenstein, line 46: Why does this show a green status when the test is `NonEnabled`?

service/src/main/java/bio/terra/catalog/common/DaoKeyHolder.java
Abulu123: this is because timestamp can never produce a null value

service/src/main/java/bio/terra/catalog/common/DaoKeyHolder.java
Abulu123: what does inline mean?

service/src/test/java/bio/terra/catalog/common/StorageSystemTest.java
Abulu123: The reason there is none is because that case is covered in other tests. Should I still add it?

service/src/test/java/bio/terra/catalog/service/BaseStatusServiceTest.java
Abulu123: well in the method weather or not the status is enabled or not is still returns a green status

service/src/main/java/bio/terra/catalog/common/DaoKeyHolder.java
s-rubenstein: Instead of doing:
```
Map<String, Object> keys = getKeys();
Object fieldObject = keys.get(fieldName);
```
Do:
```
Object fieldObject = getKeys().get(fieldName);
```

service/src/test/java/bio/terra/catalog/common/StorageSystemTest.java
s-rubenstein: If it is already covered, that's fine.

s-rubenstein (approved):
LGTM, pending the inline change.

service/src/main/java/bio/terra/catalog/common/DaoKeyHolder.java
pshapiro4broad: Yes, the column has a `nullable: false` constraint.

service/src/test/java/bio/terra/catalog/service/BaseStatusServiceTest.java
```diff
@@ -23,4 +26,23 @@ void getCurrentStatus() {
...
         is(new SystemStatus().ok(true).systems(Map.of("test", status))));
   }
+
+  interface Status extends Supplier<SystemStatusSystems> {}
+
+  @Test
+  void startStatusChecking() {
+    var config = new StatusCheckConfiguration(true, 1, 0, 10);
+    BaseStatusService service = new BaseStatusService(config);
+    var status = mock(Status.class);
```
tcjiang9, line 36: can you use @mock here?

service/src/test/java/bio/terra/catalog/service/BaseStatusServiceTest.java
pshapiro4broad: It returns `true`, although I wonder if it should? Maybe we should be returning `false` instead. I'm not even sure why the configuration attribute `enabled` exists in the first place, it's always `true` in all our configurations.

tcjiang9 (approved):

service/src/main/java/bio/terra/catalog/service/BaseStatusService.java
```diff
@@ -57,16 +58,12 @@ void registerStatusCheck(String name, Supplier<SystemStatusSystems> checkFn) {
   void checkStatus() {
     if (configuration.enabled()) {
       var newStatus = new SystemStatus();
-      try {
-        var systems =
-            statusCheckMap.entrySet().stream()
-                .collect(Collectors.toMap(Map.Entry::getKey, e -> e.getValue().get()));
-        newStatus.setOk(systems.values().stream().allMatch(SystemStatusSystems::isOk));
-        newStatus.setSystems(systems);
-      } catch (Exception e) {
```
pshapiro4broad, line 66: I think it's incorrect to remove this code. The try/catch is here to handle the case where a subsystem throws an exception when its status check method is called. When a status check throws an exception we want to capture that error and show that the subsystem is down. You can test this case using code like:
```
  void getCurrentStatusException() {
    var config = new StatusCheckConfiguration(true, 0, 0, 10);
    BaseStatusService service = new BaseStatusService(config);
    var status = new SystemStatusSystems().ok(true);
    service.registerStatusCheck("test", () -> {throw new RuntimeException("failure");});
    SystemStatus systemDown = new SystemStatus().ok(false);
    assertThat(service.getCurrentStatus(), is(systemDown));
    service.checkStatus();
    assertThat(service.getCurrentStatus(), is(systemDown));
  }
```

service/src/test/java/bio/terra/catalog/service/BaseStatusServiceTest.java
pshapiro4broad: You could, but you'd have to also annotate the class with `@ExtendWith(MockitoExtension.class)`. I think it's OK to call `mock()` directly since it's only used in one test rather than annotating the class. But either way would work.

pshapiro4broad (changes requested):
One thing to change, I think the try/catch in `checkStatus()` is required and can't be removed. Also maybe we should remove `status-check.enabled` as a application property, unless there's a use case for it.

service/src/test/java/bio/terra/catalog/service/BaseStatusServiceTest.java
pshapiro4broad: Resolving this is probably out of scope for this PR. Keep the test you have and create a tech debt ticket in JIRA so we can look at this later.

pshapiro4broad (approved):

PR #148, opened 2022-11-09
Make changes to signature for create dataset, and minor changes to sc…

s-rubenstein:
…hema to create datasets

### DIFF
--- a/service/src/main/java/bio/terra/catalog/service/DatasetService.java
+++ b/service/src/main/java/bio/terra/catalog/service/DatasetService.java
@@ -196,11 +196,16 @@ public class DatasetService {
   }
 
   public DatasetId createDataset(
-      StorageSystem storageSystem, String storageSourceId, String metadata) {
-    jsonValidationService.validateMetadata(toJsonNode(metadata));
-    var dataset = new Dataset(storageSourceId, storageSystem, metadata);
-    ensureActionPermission(dataset, SamAction.CREATE_METADATA);
-    return datasetDao.create(dataset).id();
+      StorageSystem storageSystem, String storageSourceId, Map<String, Object> metadata) {
+    try {
+      String metadataAsString = objectMapper.writeValueAsString(metadata);
+      jsonValidationService.validateMetadata(toJsonNode(metadataAsString));
+      var dataset = new Dataset(storageSourceId, storageSystem, metadataAsString);
+      ensureActionPermission(dataset, SamAction.CREATE_METADATA);
+      return datasetDao.create(dataset).id();
+    } catch (JsonProcessingException e) {
+      throw new BadRequestException("Metadata is not an object", e);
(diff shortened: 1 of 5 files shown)


service/src/test/java/bio/terra/catalog/service/DatasetServiceTest.java
```diff
@@ -270,25 +272,13 @@ void testCreateDatasetWithInvalidUser() {
...
-    when(datasetDao.create(new Dataset(SOURCE_ID, dataset.storageSystem(), METADATA)))
+    when(datasetDao.create(new Dataset(SOURCE_ID, dataset.storageSystem(), METADATA_AS_STRING)))
         .thenReturn(dataset);
     DatasetId id = datasetService.createDataset(dataset.storageSystem(), SOURCE_ID, METADATA);
     verify(jsonValidationService).validateMetadata(objectMapper.readTree(dataset.metadata()));
     assertThat(id, is(datasetId));
   }
 
-  @Test
-  void testCreateDatasetInvalidMetadata() {
```
s-rubenstein, line 281: We don't need this test anymore, because typing will take care of this before it gets to this point now!

service/src/main/resources/schema/development/schema.json
```diff
@@ -3,7 +3,7 @@
   "$id": "https://app.terra.bio/schemas/datacatalog.development.json",
   "title": "catalog dataset",
   "type": "object",
-  "required": ["dct:title", "dct:description", "dct:creator", "dct:issued", "TerraDCAT_ap:hasDataCollection", "dcat:accessURL", "prov:wasGeneratedBy", "storage", "counts", "contributors"],
```
s-rubenstein, line 6: I stole this change from @tcjiang9 's PR, because it is necessary to create Nate's datasets.

service/src/main/java/bio/terra/catalog/service/DatasetService.java
```diff
@@ -196,11 +196,16 @@ public void updateMetadata(DatasetId datasetId, String metadata) {
...
 
   public DatasetId createDataset(
-      StorageSystem storageSystem, String storageSourceId, String metadata) {
-    jsonValidationService.validateMetadata(toJsonNode(metadata));
-    var dataset = new Dataset(storageSourceId, storageSystem, metadata);
-    ensureActionPermission(dataset, SamAction.CREATE_METADATA);
-    return datasetDao.create(dataset).id();
+      StorageSystem storageSystem, String storageSourceId, Map<String, Object> metadata) {
+    try {
+      String metadataAsString = objectMapper.writeValueAsString(metadata);
```
s-rubenstein, line 201: I think this should be a service level change, because mapping to a string for internal storage is business logic. My belief is that controller code should purely be used for HTTP validation.

PR #149, opened 2022-11-10
DC-574: Look at how we return 401 vs 403 errors and fix if necessary

Abulu123:
changed the `UnauthorizedException` to a `ForbiddenException` in the `ensureActionPermission` method


pshapiro4broad (approved):
Looks good. I think you will also need to update the integration tests. Look for the use of `STATUS_CODE_UNAUTHORIZED`.

tcjiang9 (approved):

PR #150, opened 2022-11-10
[DC-412][DC-547] Fix getDataset and change create to upsert

s-rubenstein:


s-rubenstein:
The missing test is on the exception, and I would advocate for not testing that, since its just passing through to TCL

service/src/main/java/bio/terra/catalog/service/ExternalSystemService.java
```diff
@@ -32,6 +32,10 @@ public Map<String, StorageSystemInformation> getDatasets() {
                 dataset -> new StorageSystemInformation(DatasetAccessLevel.DISCOVERER, null)));
   }
 
+  public StorageSystemInformation getDataset(String storageSourceId) {
+    return new StorageSystemInformation(getRole(storageSourceId), null);
```
pshapiro4broad, line 36: This form of calling `StorageSystemInformation(x, null)` is used in two places now. Can you add a constructor override that only accepts a single parameter for this usage? Looking at all usages of this object's constructor I see two cases:
- both access level and phs id
- only access level

The current definition uses a no-arg constructor to support a builder-like pattern. But since it's only ever used in these two ways, I think it would be better to remove the builder pattern for now, and add a custom constructor for the `only access level` case. The other case is already handled by the `record` provided constructor.

service/src/test/java/bio/terra/catalog/service/DatasetServiceTest.java
```diff
@@ -129,6 +130,16 @@ void listDatasets() {
     assertThat(tdrJson.get("accessLevel").asText(), is(String.valueOf(DatasetAccessLevel.OWNER)));
   }
 
+  @Test
+  void getMetadata() throws JSONException {
```
pshapiro4broad, line 134: Minor nit, I would use `throws Exception` for tests since there's (IMO) nothing to be gained by using a more specific exception type, and it requires code to be changed if the thrown exception changes.

service/src/test/java/bio/terra/catalog/service/DatasetServiceTest.java
```diff
@@ -129,6 +130,16 @@ void listDatasets() {
...
+    when(datarepoService.getRole(SOURCE_ID)).thenReturn(DatasetAccessLevel.OWNER);
+    when(datarepoService.getDataset(SOURCE_ID))
+        .thenReturn(new StorageSystemInformation().datasetAccessLevel(DatasetAccessLevel.OWNER));
+    when(datasetDao.retrieve(any())).thenReturn(tdrDataset);
```
pshapiro4broad, line 138: Shouldn't this be
```suggestion
    when(datasetDao.retrieve(tdrDataset.id())).thenReturn(tdrDataset);
```

Also, is there a reason to use the TDR service here? Why not use the external service, which would let you use `mockDataset()`?

I think overall this test code should be overhauled to use the same service for all tests (which can be any service at this point, since at this layer no service has any special behavior) except for a test that's specifically testing the mapping of storage system type to service, or of handling multiple services at the same time (such as in `listDatasets()`).

service/src/test/java/bio/terra/catalog/service/dataset/DatasetDaoTest.java
```diff
@@ -31,7 +30,12 @@ class DatasetDaoTest {
           {"sampleId": "12345", "species": ["mouse", "human"]}""";
 
   private Dataset createDataset(String storageSourceId, StorageSystem storageSystem) {
```
pshapiro4broad, line 32: maybe rename these to `upsert` as well?

service/src/main/java/bio/terra/catalog/service/dataset/exception/InvalidDatasetException.java
```diff
@@ -1,9 +1,9 @@
 package bio.terra.catalog.service.dataset.exception;
 
-import bio.terra.common.exception.BadRequestException;
+import bio.terra.common.exception.ConflictException;
 
-public class InvalidDatasetException extends BadRequestException {
-  public InvalidDatasetException(String message, Throwable cause) {
-    super(message, cause);
+public class InvalidDatasetException extends ConflictException {
```
pshapiro4broad, line 5: Is this exception still thrown? I don't see where it's created.

pshapiro4broad (approved):
Looks good, I have a few comments.

integration/src/main/java/scripts/testscripts/DatasetPermissionOperations.java
```diff
@@ -172,7 +172,8 @@ private void testDiscovererPermissionsForSnapshotDataset() throws Exception {
     // TODO (DR-2833): Uncomment once TDR makes retrieve snapshot work for discoverers.
```
pshapiro4broad, line 172: To avoid checking in commented out code, you could instead move this code to the JIRA ticket.

s-rubenstein: I'll make a corresponding ticket for our board.

petesantos (approved):

PR #151, opened 2022-11-17
[DC-420] Multithread storage system dataset fetching

s-rubenstein:
Before:
curl -X 'GET' 'localhost:8080/api/v1/datasets' -H 'accept: application/json'   0.01s user 0.01s system 1% cpu 1.451 total

After:
curl -X 'GET' 'localhost:8080/api/v1/datasets' -H 'accept: application/json'   0.01s user 0.02s system 7% cpu 0.292 total

This is round one of performance improvements. The next planned change is to do all the database fetching at once, rather than do a separate dao call per storage system.

### DIFF
--- a/service/src/main/java/bio/terra/catalog/service/DatasetService.java
+++ b/service/src/main/java/bio/terra/catalog/service/DatasetService.java
@@ -24,6 +24,9 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
 import org.springframework.stereotype.Service;
 
 @Service
@@ -132,11 +135,14 @@ public class DatasetService {
     return createDatasetResponses(roleMap, system);
   }
 
-  public DatasetsListResponse listDatasets() {
+  public DatasetsListResponse listDatasets() throws InterruptedException {
     List<DatasetResponse> datasets = new ArrayList<>();
+    ExecutorService es = Executors.newCachedThreadPool();
     for (StorageSystem system : StorageSystem.values()) {
-      datasets.addAll(collectDatasets(system));
+      es.execute(() -> datasets.addAll(collectDatasets(system)));
     }
+    es.shutdown();
+    es.awaitTermination(1, TimeUnit.MINUTES);
     if (samService.hasGlobalAction(SamAction.READ_ANY_METADATA)) {
       datasets.addAll(
           datasetDao.listAllDatasets().stream()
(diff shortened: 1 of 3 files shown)


service/src/main/java/bio/terra/catalog/service/DatasetService.java
```diff
@@ -132,11 +135,14 @@ private List<DatasetResponse> collectDatasets(StorageSystem system) {
     return createDatasetResponses(roleMap, system);
   }
 
-  public DatasetsListResponse listDatasets() {
+  public DatasetsListResponse listDatasets() throws InterruptedException {
```
pshapiro4broad, line 138: I think we should keep the current convention that these APIs don't throw checked exceptions. This could catch `Interrupted` and rethrow another exception instead. (Although note that sonar will want you to interrupt the current thread when you catch `InterruptedException`.)

service/src/main/java/bio/terra/catalog/service/DatasetService.java
```diff
@@ -132,11 +135,14 @@ private List<DatasetResponse> collectDatasets(StorageSystem system) {
...
     List<DatasetResponse> datasets = new ArrayList<>();
+    ExecutorService es = Executors.newCachedThreadPool();
     for (StorageSystem system : StorageSystem.values()) {
-      datasets.addAll(collectDatasets(system));
+      es.execute(() -> datasets.addAll(collectDatasets(system)));
```
pshapiro4broad, line 142: I'm surprised that this works. So the current user token, which is request scoped, is propagated to the other threads correctly?

service/src/main/java/bio/terra/catalog/service/DatasetService.java
```diff
@@ -132,11 +135,14 @@ private List<DatasetResponse> collectDatasets(StorageSystem system) {
     return createDatasetResponses(roleMap, system);
   }
 
-  public DatasetsListResponse listDatasets() {
+  public DatasetsListResponse listDatasets() throws InterruptedException {
     List<DatasetResponse> datasets = new ArrayList<>();
+    ExecutorService es = Executors.newCachedThreadPool();
```
pshapiro4broad, line 140: Can you use a parallel stream instead of creating an executor service?
Something like
```
List<DatasetResponse> datasets = Arrays.stream(StorageSystem.values()).parallel().map(this::collectDatasets).toList();
```

service/src/main/java/bio/terra/catalog/service/DatasetService.java
s-rubenstein: I am not actually sure. I am poking around a bit and we will see. Its definitely getting some results, but I am not sure what it is getting that is admin, rather than from the actual fetching. Going to comment out the admin code to see.

s-rubenstein: Yeah, it doesn't work. I'll poke around.

PR #152, opened 2022-11-21
multithread storage systems

pshapiro4broad:
Making a draft PR to get tests running

### DIFF
--- a/service/src/main/java/bio/terra/catalog/common/RequestContextCopier.java
+++ b/service/src/main/java/bio/terra/catalog/common/RequestContextCopier.java
@@ -1,0 +1,58 @@
+package bio.terra.catalog.common;
+
+import org.springframework.web.context.request.RequestAttributes;
+import org.springframework.web.context.request.RequestContextHolder;
+
+import java.util.stream.Stream;
+
+/**
+ * This class supports creating a parallel {@code Stream} with a copy of the current thread's
+ * Spring request attributes copied to the stream's sub-threads. This has the effect of
+ * allowing request scope injected fields to be used in classes that are passed to the sub-threads.
+ * <p/>
+ * For example, if {@code service} uses a request scope injection, and you want to run it in
+ * parallel, you can write:
+ *
+ * <pre>{@code
+ *     int sum = RequestContextCopier.parallelWithRequest(widgets.stream())
+ *                      .mapToInt(w -> service.getWeight(w))
+ *                      .sum();
+ * }</pre>
+ *
+ * @see RequestContextHolder
+ * @see RequestAttributes
+ * @see Stream#parallel()
+ */
+public class RequestContextCopier {
+
+  private final RequestAttributes attributes = RequestContextHolder.getRequestAttributes();
+
+  private RequestContextCopier() {
+  }
+
+  private <T> void copyToThread(T target) {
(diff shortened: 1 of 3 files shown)


service/src/main/java/bio/terra/catalog/service/DatasetService.java
```diff
@@ -135,21 +136,32 @@ private List<DatasetResponse> collectDatasets(StorageSystem system) {
     return createDatasetResponses(roleMap, system);
   }
 
+  static class ContextCopier {
```
s-rubenstein, line 139: I would call this something around `ParallelRequestWithHttpContextCopying` or something? My only concern is remembering the _purpose_ of the class when we come back to it. (I know there's the comment on the method, but I think on the class would be good.

s-rubenstein (approved):
Overall, I like this change. I worry about the understandability of it down the line, but I think renaming could help with that.

service/src/main/java/bio/terra/catalog/service/DatasetService.java
pshapiro4broad: I'll put it out to its own class and will try a different name there.

I'd like to find a shorter name than `ParallelRequestWithHttpContextCopying`. I think `http` is not needed since the request context can contain other things besides http state. And the "parallel with" is part of the operation itself I think. Maybe `RequestContextCopier.parallelWithContext()`? Although that says `Context` twice which seems redundant too...

s-rubenstein: If you really want to avoid `Parallel` being in there, I think that this is probably the best name. But yeah, the question in my mind is whether we will understand what this class is for in the future. Because it is just used in this class, it could potentially be a static method in this class, rather than a new class. That may be clearer in this case, in my opinion, just because my confusion is mostly around the intermediate naming for the class.

service/src/main/java/bio/terra/catalog/common/RequestContextCopier.java
```diff
@@ -0,0 +1,58 @@
...
+   * using {@link Stream#parallel()}.
+   *
+   * @param stream the stream to parallelize
+   * @return the new parallel stream
+   * @param <S> the type of the stream elements
+   */
+  // This usage of `peek()` is correct, even though it generates a sonar warning. If it's never
+  // called (because the element in the stream is never consumed), that's fine.
+  @SuppressWarnings("java:S3864")
+  public static <S> Stream<S> parallelWithRequest(Stream<S> stream) {
```
pshapiro4broad, line 51: Since this is inside `RequestContextCopier` it could be called `parallel()` (or `createParallel()`), to avoid repeating `Request`?

service/src/main/java/bio/terra/catalog/common/RequestContextCopier.java
```diff
@@ -0,0 +1,58 @@
+package bio.terra.catalog.common;
+
+import org.springframework.web.context.request.RequestAttributes;
+import org.springframework.web.context.request.RequestContextHolder;
+
+import java.util.stream.Stream;
+
+/**
+ * This class supports creating a parallel {@code Stream} with a copy of the current thread's
```
s-rubenstein, line 9: Great! This comment adds what I was hoping for.

PR #153, opened 2022-11-22
[DC-420] Sr multithread storage systems

s-rubenstein:
Loading the dataset list from the catalog can take a long time. One of the reasons for this is because we were loading the data from underlying storage systems in sequence. This PR moves that to loading the underlying data in parallel, and then unifying once all the data has been fetched. It does this by using stream parallelism. The added complexity in this PR is a result of Spring Request Context not being copied down to child threads. In this case, because we know this request doesn't close until after all the child threads have finished execution, we want to copy down that request context.

Ticket: https://broadworkbench.atlassian.net/browse/DC-420

Timing:
Before:
curl -X 'GET' 'localhost:8080/api/v1/datasets' -H 'accept: application/json'   0.01s user 0.01s system 1% cpu 1.451 total
After:
curl -X 'GET' 'http://localhost:8080/api/v1/datasets' -H  -H   0.01s user 0.01s system 3% cpu 0.445 total

### DIFF
--- a/service/src/test/java/bio/terra/catalog/common/RequestContextCopierTest.java
+++ b/service/src/test/java/bio/terra/catalog/common/RequestContextCopierTest.java
@@ -33,7 +33,9 @@ class RequestContextCopierTest {
 
   @Test
   @Disabled
-  // This is disabled because the way in which Java optimizes parallelization is tricky to test in junit. It works consistently in the positive test, but in the negative test it sometimes optimizes things away where we would like it not to.
+  // This is disabled because the way in which Java optimizes parallelization is tricky to test in
+  // junit. It works consistently in the positive test, but in the negative test it sometimes
+  // optimizes things away where we would like it not to.
   void requestContextDoesntFollowToChildThreads() {
     assertThrows(IllegalStateException.class, () -> runTestWithProvider(Stream::parallel));
   }
--- a/service/src/test/java/bio/terra/catalog/common/RequestContextCopierTest.java
+++ b/service/src/test/java/bio/terra/catalog/common/RequestContextCopierTest.java
@@ -1,0 +1,29 @@
+package bio.terra.catalog.common;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.util.Objects;
+import java.util.stream.Stream;
+import org.junit.jupiter.api.Test;
+import org.springframework.mock.web.MockHttpServletRequest;
+import org.springframework.web.context.request.RequestAttributes;
+import org.springframework.web.context.request.RequestContextHolder;
+import org.springframework.web.context.request.ServletRequestAttributes;
+
+public class RequestContextCopierTest {
+  @Test
+  void setsRequestContextInsideThread() {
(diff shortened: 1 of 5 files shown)


service/src/test/java/bio/terra/catalog/common/RequestContextCopierTest.java
```diff
@@ -0,0 +1,27 @@
...
+import org.springframework.web.context.request.ServletRequestAttributes;
+
+public class RequestContextCopierTest {
+  @Test
+  void setsRequestContextInsideThread() {
+    MockHttpServletRequest request = new MockHttpServletRequest();
+    request.setAttribute("attribute", "value");
+    RequestContextHolder.setRequestAttributes(new ServletRequestAttributes(request));
+    RequestContextCopier.parallelWithRequest(Stream.of(1, 2))
+        .forEach(
```
pshapiro4broad, line 20: I was wondering how you were going to test this!

Can you add a negative test too?

Also, using `forEach(..assertEquals())` could result in a false positive since it will succeed even if `forEach()` never runs. One way around this would be to use a terminal operation like `sum()` and then assert that the terminal state of the stream is correct as well, or have the `map()` consume the request scope state and use it for its operation.

pshapiro4broad (approved):
Looks good, just some suggestions on the test code.

service/src/test/java/bio/terra/catalog/common/RequestContextCopierTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+  @Test
+  void setsRequestContextInsideThread() {
+    assertEquals(THREADS, runTestWithProvider(RequestContextCopier::parallelWithRequest));
+  }
+
+  @Test
+  @Disabled
+  void requestContextDoesntFollowToChildThreads() {
+    assertThrows(IllegalStateException.class, () -> runTestWithProvider(Stream::parallel));
```
s-rubenstein, line 37: This is disabled because the way in which Java optimizes parallelization is tricky to test in junit. It works consistently in the positive test, but in the negative test it sometimes optimizes things away where we would like it not to.

pshapiro4broad: To make sonar happy, could you add a comment to the test that has this info?
https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&pullRequest=153&id=terra-data-catalog&open=AYTEt4YSxq4-Lt3XYr3u

Also it wants you to remove `public` from the `class` declaration.

petesantos (commented):
This looks good, a few minor (non-code) comments:
1. Could you add a short description of the problem and solution to the PR? I find it easier to understand the PR when the description is close to the change if I were to look through the repo at a later date
2. Were you able to test the performance before and after? I am curious what the benefit looks like

petesantos (approved):

okotsopoulos (approved):

okotsopoulos:
I will second @petesantos' comment -- the ticket itself has great context but I also find it useful to have it close to the changes, along with a link to the ticket in question.

PR #154, opened 2022-11-22
DC-594: list dataset performance improvement: reduce DB calls

pshapiro4broad:
Split out database operation from code that collects datasets from systems, to avoid doing multiple database calls.

The query builder code isn't great, but it is careful to not do any text substitution. It only builds a string that contains `?` placeholders that JdbcTemplate uses to do the actual parameter substitution.

Note that this PR conflicts with with https://github.com/DataBiosphere/terra-data-catalog/pull/154 and will be merged after that PR is merged.

### DIFF
--- a/service/src/main/java/bio/terra/catalog/service/DatasetService.java
+++ b/service/src/main/java/bio/terra/catalog/service/DatasetService.java
@@ -20,10 +20,11 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.fasterxml.jackson.databind.node.TextNode;
-import java.util.HashMap;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 import org.springframework.stereotype.Service;
 
@@ -115,11 +116,11 @@ public class DatasetService {
   }
 
   public DatasetsListResponse listDatasets() {
-    Map<StorageSystem, Map<String, StorageSystemInformation>> systemsAndInfo = new HashMap<>();
-    for (StorageSystem system : StorageSystem.values()) {
-      var systemDatasets = getService(system).getDatasets();
-      systemsAndInfo.put(system, systemDatasets);
-    }
+    var systemsAndInfo =
+        Arrays.stream(StorageSystem.values())
+            .collect(
+                Collectors.toMap(Function.identity(), system -> getService(system).getDatasets()));
+
     List<Dataset> datasets;
     if (samService.hasGlobalAction(SamAction.READ_ANY_METADATA)) {
       datasets = datasetDao.listAllDatasets();
@@ -131,8 +132,9 @@ public class DatasetService {
                       Collectors.toMap(Map.Entry::getKey, entry -> entry.getValue().keySet())));
     }
     var response = new DatasetsListResponse();
-    var defaultInformation =
-        new StorageSystemInformation(DatasetAccessLevel.READER);
+    // This is used for the admin user, to provide dummy role for datasets an admin
+    // doesn't have access to in the underlying storage system.
+    var defaultInformation = new StorageSystemInformation(DatasetAccessLevel.READER);
     response.setResult(
         datasets.stream()
             .map(
--- a/service/src/main/java/bio/terra/catalog/service/DatasetService.java
+++ b/service/src/main/java/bio/terra/catalog/service/DatasetService.java
@@ -133,7 +133,7 @@ public class DatasetService {
                       Collectors.toMap(Map.Entry::getKey, entry -> entry.getValue().keySet())));
     }
     var response = new DatasetsListResponse();
-    // This is used for the admin user, to provide dummy role for datasets an admin
+    // This is used for the admin user, to provide a dummy role for datasets that an admin
     // doesn't have access to in the underlying storage system.
     var defaultInformation = new StorageSystemInformation(DatasetAccessLevel.READER);
     response.setResult(
--- a/service/src/main/java/bio/terra/catalog/service/DatasetService.java
+++ b/service/src/main/java/bio/terra/catalog/service/DatasetService.java
@@ -122,7 +122,6 @@ public class DatasetService {
             .collect(
                 Collectors.toMap(Function.identity(), system -> getService(system).getDatasets()));
 
-
     List<Dataset> datasets;
     if (samService.hasGlobalAction(SamAction.READ_ANY_METADATA)) {
       datasets = datasetDao.listAllDatasets();
--- a/service/src/main/java/bio/terra/catalog/service/DatasetService.java
+++ b/service/src/main/java/bio/terra/catalog/service/DatasetService.java
@@ -20,10 +20,11 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.fasterxml.jackson.databind.node.TextNode;
-import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
+import java.util.stream.Collectors;
 import org.springframework.stereotype.Service;
 
 @Service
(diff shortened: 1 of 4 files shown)


service/src/main/java/bio/terra/catalog/service/DatasetService.java
```diff
@@ -113,48 +115,37 @@ private ObjectNode toJsonNode(String json) {
...
+              systemsAndInfo.entrySet().stream()
+                  .collect(
+                      Collectors.toMap(Map.Entry::getKey, entry -> entry.getValue().keySet())));
     }
     var response = new DatasetsListResponse();
-
-    response.setResult(datasets.stream().map(DatasetResponse::convertToObject).toList());
+    // This is used for the admin user, to provide dummy role for datasets an admin
+    // doesn't have access to in the underlying storage system.
+    var defaultInformation = new StorageSystemInformation(DatasetAccessLevel.READER);
```
s-rubenstein, line 137: I would just inline this below, and put the comment there. That way people don't have to follow the chain up for `getOrDefault` to see why that is being done.

service/src/main/java/bio/terra/catalog/service/dataset/DatasetDao.java
```diff
@@ -76,7 +79,7 @@ public Dataset upsert(Dataset dataset) {
   }
 
   @WriteTransaction
-  public Dataset update(Dataset dataset) {
```
s-rubenstein, line 79: Why this change? It seems unrelated to me?

service/src/main/java/bio/terra/catalog/service/dataset/DatasetDao.java
```diff
@@ -76,7 +79,7 @@ public Dataset upsert(Dataset dataset) {
...
+  public void update(Dataset dataset) {
```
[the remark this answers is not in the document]
pshapiro4broad: @s-rubenstein

service/src/test/java/bio/terra/catalog/service/DatasetServiceTest.java
```diff
@@ -173,12 +176,9 @@ void listDatasetsUsingAdminPermissions() {
     when(datarepoService.getDatasets()).thenReturn(datasets);
     when(rawlsService.getDatasets()).thenReturn(workspaces);
     when(samService.hasGlobalAction(SamAction.READ_ANY_METADATA)).thenReturn(true);
-    when(datasetDao.find(StorageSystem.TERRA_WORKSPACE, workspaces.keySet())).thenReturn(List.of());
-    when(datasetDao.find(StorageSystem.TERRA_DATA_REPO, datasets.keySet()))
-        .thenReturn(List.of(tdrDataset));
     when(datasetDao.listAllDatasets()).thenReturn(List.of(workspaceDataset, tdrDataset));
-    ObjectNode tdrJson = (ObjectNode) datasetService.listDatasets().getResult().get(0);
-    ObjectNode workspaceJson = (ObjectNode) datasetService.listDatasets().getResult().get(1);
+    ObjectNode tdrJson = (ObjectNode) datasetService.listDatasets().getResult().get(1);
```
pshapiro4broad, line 180: This order dependency is ugly, I'm looking at ways to fix this using the hamcrest json object matchers.

service/src/main/java/bio/terra/catalog/service/DatasetService.java
pshapiro4broad: I checked and inlining the call does cause multiple `StorageSystemInformation`s to be created. If there was a JVM defaulting map like python has, that would be a way to write this and avoid the duplicate object creation. Many third party options exist (like Apache's `DefaultedMap`) but I'd like to avoid third party libraries if there's a JVM equivalent form, and `Map.getOrDefault()` is pretty close.

pshapiro4broad: To be clear; the main problem is that this object is constructed many times regardless of if it's needed or not. Since Java creates all arguments before calling a function, calling `Map.getOrDefault(x, function())` will invoke `function()` even if the default value it generates is never used. Other methods like `Objects.requireNonNull()` solve this problem by providing an API that accepts `Supplier<T>` as well as `T` so you can put a lambda there and  it won't be called unless the value is needed.

pshapiro4broad: Relevant stack overflow thread about this https://stackoverflow.com/a/64486585/947286

service/src/main/java/bio/terra/catalog/service/dataset/DatasetDao.java
pshapiro4broad: It is unrelated. It was changed due to a sonar / IntelliJ warning because the method returns a value but the return value is never used. If at some point the return value is used, it can be added back.

service/src/main/java/bio/terra/catalog/service/DatasetService.java
pshapiro4broad: Another approach would be to add all missing entries to the map so `getOrDefault()` is no longer needed. Storage-wise this is wasteful but it only affects the admin user case and would simplify this code. I'll look at doing it this way and see how it turns out.

pshapiro4broad: I tried this and couldn't figure out a way to do it in less / clearer code than is here already.

service/src/main/java/bio/terra/catalog/service/DatasetService.java
```diff
@@ -115,49 +116,37 @@ private ObjectNode toJsonNode(String json) {
...
+    response.setResult(
+        datasets.stream()
+            .map(
+                dataset ->
+                    new DatasetResponse(
+                        dataset,
+                        systemsAndInfo
+                            .getOrDefault(dataset.storageSystem(), Map.of())
```
s-rubenstein, line 146: TOL: I wonder if it would be worth adding an 'UNKNOWN' storageSystem. Not for this PR, but maybe a ticket if we think it is a good idea.

s-rubenstein (approved):
LGTM

service/src/main/java/bio/terra/catalog/service/DatasetService.java
pshapiro4broad: I'd be interested to hear more about how we would use it. I think unknown objects are good for cases where you want to avoid `null` checks and to provide default behavior. But since `getOrDefault()` is needed to handle the admin user case I'm not sure how an unknown storage system would be used here.

s-rubenstein: Yeah! Of course! I was thinking it might make it easier for clients to use accessLevel if they can assume it is a non-nullable field (even for admins). That seems like a very small use though, and not necessarily worth the time to add.

service/src/main/java/bio/terra/catalog/service/dataset/DatasetDao.java
```diff
@@ -112,15 +115,33 @@ public Dataset mapRow(ResultSet rs, int rowNum) throws SQLException {
...
+  public List<Dataset> find(Map<StorageSystem, Collection<String>> systemsAndIds) {
+    String query = "(storage_system = ? AND storage_source_id IN (%s))";
+    List<Object> args = new ArrayList<>();
+    String whereClause =
+        systemsAndIds.entrySet().stream()
+            .filter(entry -> !entry.getValue().isEmpty())
+            .map(
+                entry -> {
+                  args.add(String.valueOf(entry.getKey()));
+                  args.addAll(entry.getValue());
```
petesantos, line 130: nit: Is there any way we could cleanly separate the arg building from the map? The rationale being the map function has a side effect of building the arguments.

I was thinking of something like having the map return tuples of args and clauses, but I think that would add too much complexity...

I know you mentioned you weren't happy with the query building, and this isn't critical.

petesantos (approved):
LGTM

service/src/main/java/bio/terra/catalog/service/dataset/DatasetDao.java
pshapiro4broad: I thought about this; having one method that builds the query string and another that populates the args map. The main problems with doing it separately is that you end up duplicating the the `stream()` operation and it's less clear (to me at least) that the query string is mirroring the arg list. On the other hand, doing it separately means that it's more obvious that the query only contains `?`s and not any other data.

I did look a bit but didn't see anything that would help here, other than using a higher level query language to write our DAO in. Spring supports a few ways to do this, but that would be a pretty big change to the DAO code. It might be worthwhile looking into at some point though, as our DAO is pretty simple so far.

PR #156, opened 2022-11-30
DC-524: Add script to validate metadata in the catalog database

pshapiro4broad:
Implement a validation script for the catalog as a spring command line runner.

This PR restructures the code in the repo so that most of the spring services are now in a new subproject, `common`. The `service` subproject now only contains the controller classes and other web API deployment components.  Another new subproject, `admin-cli`, contains the new CLI program.

See instructions on `admin-cli` in [admin-cli/README-cli.md](https://github.com/DataBiosphere/terra-data-catalog/blob/8841d6c4c639c5f904ba2a256583ef36d691b8db/admin-cli/README-cli.md)

Still to do:
- ~clean up gradle configs and application.yaml for subprojects and remove unnecessary content~
- ~look into running sonar at the root level of the repo to scan `admin-cli` and `service` subprojects too~ This generated an out of memory error in sonar. I'll create a ticket to revisit this later. The bulk of the catalog code is in `common`.
- ~maybe rename `service` subproject to `web-api`~ This may be a good idea, but is not directly related to this change and will make the PR bigger. I'll make a ticket for this too.


service/src/main/resources/application.yml
```diff
@@ -21,12 +21,6 @@ env:
 
 # Below here is non-deployment-specific
 
-# When the target is 'local' the write-config.sh script will generate this properties file. It
-# contains the configuration of the BPM test application. We can use that application in our
-# integration testing to make sure the application code paths are working. However, we do not
-# want it to appear in production environments.
-spring.config.import: optional:file:../config/local-properties.yml
```
pshapiro4broad, line 28: This change isn't directly related, but I noticed it and we don't use `local-properties.yml`. The `write-config.sh` script it refers to doesn't exist in this project either.

service/src/main/resources/application.yml
```diff
@@ -44,7 +38,7 @@ spring:
           maxAge: 0
           mustRevalidate: true
         useLastModified: false
-      staticLocations: classpath:/api/
+      staticLocations: classpath:/api/,classpath:/schema/development/
```
pshapiro4broad, line 41: This is an unrelated change I was testing out, I'll revert this.

service/src/test/java/bio/terra/catalog/controller/DatasetApiControllerTest.java
```diff
@@ -1,4 +1,4 @@
-package bio.terra.catalog.api;
+package bio.terra.catalog.controller;
```
pshapiro4broad, line 1: This is (sort of) an unrelated change. The openapi generated base classes are created in package `api`. It seems like we shouldn't put generated code in the same package as hand written code, so I moved our subclasses to `controller`.

settings.gradle
```diff
@@ -1,4 +1,4 @@
 rootProject.name = 'terra-data-catalog'
-include('service', 'client', 'rawls-client', 'integration', 'scripts')
+include('admin-cli', 'service', 'client', 'rawls-client', 'library', 'integration', 'scripts')
```
pshapiro4broad, line 2: I called the subproject `library` but it occurs to me that it's not a library we build. Maybe it should be called `common` instead?

.github/workflows/integration-tests.yml
```diff
@@ -30,8 +30,8 @@ jobs:
           distribution: 'temurin'
           cache: 'gradle'
 
-      - name: Gradle build service
-        run: ./gradlew --build-cache :service:build -x test
+      - name: Build without running tests
```
s-rubenstein, line 33: Do we have an equivalent test runner job? I see the test flag still there?

admin-cli/README-cli.md
```diff
@@ -0,0 +1,52 @@
...
+```shell
+ENV=dev PORT=5432 ./scripts/db_connect.sh
+```
+
+In the proxy output, look for the database name, username and password.
+
+3. Run the command:
+
+```shell
+DATABASE_NAME=catalog DATABASE_USER=catalog DATABASE_USER_PASSWORD=<password> \
```
s-rubenstein, line 26: I don't know where are passwords are stored offhand, but if they are on vault, could we include a vault path here? (Maybe even in a ${} so that this is easy to copy and paste?) I know I just always love when I can just copy and paste a command from a readme and it works offhand.

library/src/main/java/bio/terra/catalog/config/CatalogDatabaseConfiguration.java
```diff
@@ -43,4 +50,24 @@ public void setUpgradeOnStart(boolean upgradeOnStart) {
...
     return new JdbcTransactionManager(getDataSource());
   }
+
+  // This is a "magic bean": It supplies a method that Spring calls after the application is setup,
+  // but before the port is opened for business. That lets us do database migration and stairway
+  // initialization on a system that is otherwise fully configured. The rule of thumb is that all
+  // bean initialization should avoid database access. If there is additional database work to be
+  // done, it should happen inside this method.
+  @Bean
+  public SmartInitializingSingleton postSetupInitialization(ApplicationContext applicationContext) {
```
s-rubenstein, line 60: Out of curiosity, why did this code end up changing from the BeanConfig? I just noticed the comment is the same

settings.gradle
s-rubenstein: I like `common` more than `library`, I think.

admin-cli/src/main/java/bio/terra/catalog/cli/Main.java
```diff
@@ -0,0 +1,119 @@
...
+@Component
+public class Main {
+
+  private final DatasetDao datasetDao;
+  private final JsonValidationService jsonValidationService;
+  private final ObjectMapper objectMapper;
+
+  private final Map<String, Consumer<ApplicationArguments>> commands;
+
+  Main(
```
pshapiro4broad, line 29: I didn't create unit tests for this code yet. I'm not sure how much we need unit tests for an admin-only cli, although I can imagine we'd want unit tests for any admin operation that modifies the database.

.github/workflows/integration-tests.yml
philshap: Not sure what you're asking.. although now that I think about it, I don't know why this builds `service` at all; it should really be building the integration subproject, since this build script doesn't run the service at all. I also don't know why this calls `jib`. I'll make a JIRA to track the cleanup of this code.

admin-cli/README-cli.md
pshapiro4broad: The database name, username and password are printed out as a result of running the `db_connect.sh` script, so if you follow these steps you'd have that information already. I can add a note about that here. I could also modify the db_connect script to print out the env var setting line to cut & paste, as it currently does for the `jdbc` url and `psql` command.

I was thinking that it wouldn't be _too_ hard to write a wrapper script that smooths over this part of the process. It could run the db_connect script and capture the database information. It could find the library .jar and call `java -jar` on it, forwarding any command line arguments. Maybe that would be a future enhancement though.

.github/workflows/integration-tests.yml
pshapiro4broad: Not sure what you're asking.. although now that I think about it, I don't know why this builds `service` at all; it should really be building the integration subproject, since this build script doesn't run the service at all. I also don't know why this calls `jib`. I'll make a JIRA to track the cleanup of this code.

I think the idea here is to separate build errors from test errors by doing this, but for the integration test run there's probably not much value in doing it this way.

library/src/main/java/bio/terra/catalog/config/CatalogDatabaseConfiguration.java
pshapiro4broad: It seemed like it belonged here more than in the other place, since it relies on the configuration in this object. I may have moved it around to try to get the database tests working too. I think it would work in either place; what I'd really like to do here is remove all the database and liquibase setup code and use the spring support instead. I'll check and make a ticket for that work if it doesn't exist already.

.github/workflows/integration-tests.yml
s-rubenstein: I was just noticing the name change here and wondering what was different/why. If you have a good sense of the cleanup, that works.

admin-cli/README-cli.md
s-rubenstein: Ah, yeah. A wrapper script would probably be good, but for the time being I would add that the information is part of the output of db_connect to the readme

s-rubenstein (approved):
Overall lgtm

settings.gradle
pshapiro4broad: I just realized we already have `common` where we put `openapi.yml` and `postgres-init.sql`. The sql could be moved to `scripts`, perhaps, and maybe `openapi.yml` gets its own directory. Or we could put it in `service` or `client` and have the other use a relative path to include it.

pshapiro4broad: I moved `Postgres-init.sql` in to `scripts`. I moved `openapi.yml` into `service/.../resources`

admin-cli/build.gradle
```diff
@@ -0,0 +1,23 @@
+plugins {
+    id 'bio.terra.catalog.java-spring-conventions'
+
```
okotsopoulos, line 3: Nit - extra whitespace.

admin-cli/README-cli.md
```diff
@@ -0,0 +1,52 @@
...
+```
+
+You can also run the command without building a jar using `bootRun` and `--args`:
+
+```shell
+DATABASE_NAME=catalog DATABASE_USER=catalog DATABASE_USER_PASSWORD=<password> \
+  ./gradlew :admin-cli:bootRun --args=validate
+```
+
+## Commands supported
```
okotsopoulos, line 37: JMO but I could see this getting out of date as (hopefully) additional commands are supported.  Had you considered putting these inside a CLI `--help` option?

admin-cli/README-cli.md
```diff
@@ -0,0 +1,52 @@
+# admin-cli
+
+This is a command line tool for catalog admins to examine the contents of the catalog
```
okotsopoulos, line 3: Wording nit - when I think of "examine", my mind goes to read-only operations.  I know that all of the currently supported commands fall in this camp but you might expand in the future to support options that modify the DB.

okotsopoulos (approved):
Not too much to add as far as review comments, sorry about that.

I really appreciated your demo of this script last week… my mind goes to having something similar for TDR!  There are several admin operations that require us to connect to the DB and it would be great to have vetted wrappers for such things (e.g. clearing out stuck locks on resources).  I also like how expanding the script's offerings should be a relatively minimal lift, if you're exposing service / DAO code that exists for other purposes anyway and has been tested appropriately.

admin-cli/build.gradle
pshapiro4broad: It's totally arbitrary but I've been splitting up the versioned and non-versioned plugin references with a whitespace in-between. Although now that you point this out, I realize I should make this a non-versioned reference since it's used in multiple projects now.

admin-cli/README-cli.md
pshapiro4broad: I agree, it'd be better to rely on the tool itself to document its usage. If an incorrect command is entered you do get a short "usage" message:
```
$ java -jar admin-cli.jar xyzzy
unknown command xyzzy. Valid commands are: 'list, get, validate'
```

So there is the start of some self-documentation. As things get more complicated this help will need to improve, but I think it can wait until then.

admin-cli/README-cli.md
pshapiro4broad: Good point, I'll change the wording, and looking at the README again I think I'll edit it a bit overall.

PR #157, opened 2022-12-08
DC-570: Update swagger to reflect new schema validation

Abulu123:


okotsopoulos (commented):
Seeing a JSON schema, JSON example, and stringified JSON example all defined separately makes me concerned that a future change to the schema would not fully propagate to all of those places, leading to inconsistent documentation.

Had you considered consolidating to a single JSON schema with embedded examples, and/or referencing your example in `common.openapi.yaml`?

common/openapi.yml
```diff
@@ -106,6 +106,202 @@ paths:
           application/json:
             schema:
               type: string
+            examples:
+              jsonObject:
+                summary: A sample object
```
pshapiro4broad, line 111: I've looked into this more and I think we should be able to get `externalValue` to work. The problem is that the `index.html` that ships with swagger doesn't support it yet. See https://github.com/swagger-api/swagger-ui/issues/5433

For now, you should look at the example HTML solutions on that GitHub issue and add one to our copy of `index.html` (which is in `service/src/main/resources/templates`).

service/src/main/resources/schema/development/example.json
```diff
@@ -0,0 +1,192 @@
+{
+  "dct:title": "sint",
```
pshapiro4broad, line 2: Since it wasn't obvious how to generate this example, can you document the steps to do it somewhere, such as in a file in the `docs` directory.

pshapiro4broad (commented):
I think a little more work can be done to clean this up, see my comments.

Also, did you try creating a catalog entry using this example? You should make sure it works OK before merging it up.

service/src/main/resources/schema/development/schema.json
```diff
@@ -103,7 +103,7 @@
     },
     "contributors": {
       "type": "array",
-      "items": { "$ref": "#/defs/contributor" }
+      "items": { "$ref": "#/$defs/contributor" }
```
lu-c, line 106: Deliberate change?

lu-c (approved):
Mild approval - I do not know swagger that well, but nothing's jumping out as bad

service/src/main/resources/schema/development/schema.json
Abulu123: yes

pshapiro4broad:
I've moved these changes to https://github.com/DataBiosphere/terra-data-catalog/pull/162

PR #158, opened 2022-12-15
[DC-596] Fix issue with refs

s-rubenstein:


pshapiro4broad (approved):

pshapiro4broad:
Good to get this fix in, although I wonder if there's a way to force the schema load to fail when it's invalid like this.

petesantos (approved):

PR #159, opened 2022-12-19
DC-614: Remove helm version from test output as a temporary fix for integration test failures

pshapiro4broad:
This is a stopgap fix for DC-614 until we have a more permanent solution using a testrunner or build config change.

With this change, alpha and staging integration test builds will work again, but the helm chart versions for these tests won't be recorded in the QA dashboard.


.github/workflows/integration-tests.yml
```diff
@@ -30,34 +30,8 @@ jobs:
           distribution: 'temurin'
           cache: 'gradle'
 
-      - name: Gradle build service
-        run: ./gradlew --build-cache :service:build -x test
-
-  jib:
```
pshapiro4broad, line 36: This change isn't directly related but I noticed that we're still running the `jib` build although the output of this operation isn't used.

okotsopoulos (approved):

.github/workflows/integration-tests.yml
```diff
@@ -30,34 +30,8 @@ jobs:
...
-          GIT_SHORT_HASH=$(git rev-parse --short HEAD)
-          echo "name=${GITHUB_REPO}:${GIT_SHORT_HASH}" >> $GITHUB_OUTPUT
-
-      - name: Build image locally with jib
-        run: |
-          ./gradlew --build-cache :service:jibDockerBuild \
-            --image=${{ steps.image-name.outputs.name }} \
-            -Djib.console=plain
+      - name: Build integration subproject
+        run: ./gradlew --build-cache :integration:build
```
pshapiro4broad, line 34: Before this was building the `service` subproject; this is incorrect as the `integration` subproject doesn't depend on `service` so it never needs to be built for this workflow. The build step could be removed completely, since `gradle runTest` will build it, although leaving it separate lets us split out build errors from integration tests errors, and keep the build time out of the test time.

lu-c (approved):
The changes make sense! Thank you for all teh comments elaborating on the decision! :)

PR #160, opened 2022-12-22
DC-605: fix permission issue with admin user and dataset GET operation

pshapiro4broad:
With this change, an admin user can again `GET` any dataset in the catalog, even if they don't have access to the storage object in the underlying storage system.

### DIFF
--- a/common/src/main/java/bio/terra/catalog/service/DatasetService.java
+++ b/common/src/main/java/bio/terra/catalog/service/DatasetService.java
@@ -39,7 +39,7 @@ public class DatasetService {
    * This is used for the admin user to provide default information for datasets that an admin
    * doesn't have access to in the underlying storage system.
    */
-  private static final StorageSystemInformation ADMIN_INFORMATION =
+  private static final StorageSystemInformation DEFAULT_INFORMATION =
       new StorageSystemInformation(DatasetAccessLevel.READER);
 
   private final DatarepoService datarepoService;
@@ -150,7 +150,7 @@ public class DatasetService {
                         dataset,
                         systemsAndInfo
                             .getOrDefault(dataset.storageSystem(), Map.of())
-                            .getOrDefault(dataset.storageSourceId(), ADMIN_INFORMATION)))
+                            .getOrDefault(dataset.storageSourceId(), DEFAULT_INFORMATION)))
             .map(DatasetResponse::convertToObject)
             .toList());
     return response;
@@ -180,7 +180,7 @@ public class DatasetService {
     try {
       information = getService(dataset.storageSystem()).getDataset(dataset.storageSourceId());
     } catch (DatarepoException e) {
-      information = ADMIN_INFORMATION;
+      information = DEFAULT_INFORMATION;
     }
     return new DatasetResponse(dataset, information).convertToObject().toString();
   }
--- a/common/src/main/java/bio/terra/catalog/service/DatasetService.java
+++ b/common/src/main/java/bio/terra/catalog/service/DatasetService.java
@@ -35,8 +35,8 @@ public class DatasetService {
   public static final String PHS_ID_PROPERTY_NAME = "phsId";
 
   /**
-   *  This is used for the admin user to provide default information for datasets that an admin
-   *  doesn't have access to in the underlying storage system.
+   * This is used for the admin user to provide default information for datasets that an admin
+   * doesn't have access to in the underlying storage system.
    */
   private static final StorageSystemInformation ADMIN_INFORMATION =
       new StorageSystemInformation(DatasetAccessLevel.READER);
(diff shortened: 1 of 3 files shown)


common/src/test/java/bio/terra/catalog/service/DatasetServiceTest.java
```diff
@@ -132,11 +137,30 @@ void getMetadata() throws Exception {
...
     when(externalSystemService.getDataset(SOURCE_ID))
         .thenReturn(new StorageSystemInformation(DatasetAccessLevel.DISCOVERER));
-    when(datasetDao.retrieve(dataset.id())).thenReturn(dataset);
+    mockDataset();
     JSONAssert.assertEquals(
         metadataWithId(dataset.id()), datasetService.getMetadata(dataset.id()), true);
   }
 
+  @Test
+  void getMetadataInvalidUser() {
```
pshapiro4broad, line 146: This test wasn't changed but I wanted to organize all the `getMetadata()` tests together. Also our tests follow the convention that a test has the same name as the API it's testing, so I renamed it too.

common/src/main/java/bio/terra/catalog/service/DatasetService.java
```diff
@@ -170,10 +175,13 @@ public void deleteMetadata(DatasetId datasetId) {
...
     var dataset = datasetDao.retrieve(datasetId);
     ensureActionPermission(dataset, SamAction.READ_ANY_METADATA);
-    return new DatasetResponse(
-            dataset, getService(dataset.storageSystem()).getDataset(dataset.storageSourceId()))
-        .convertToObject()
-        .toString();
+    StorageSystemInformation information;
+    try {
+      information = getService(dataset.storageSystem()).getDataset(dataset.storageSourceId());
+    } catch (ForbiddenException e) {
```
okotsopoulos, line 181: Maybe I'm missing something, but could a non-admin caller legitimately not have access to the underlying dataset and find themselves in this catch block, and illegitimately get access that way?

pshapiro4broad: The thing to keep in mind with the catalog is that the catalog's data (a catalog entry) is separate from the underlying object's data, which a catalog entry represents. Everything in the catalog (all catalog entries) are visible to, and editable by, a catalog admin. The underlying object's data (which this code calls the storage system data) may or may not be visible to a catalog admin user.

Since the catalog itself has no special access to any data in an underlying system (like TDR), the catalog service itself can never show a user data that they would't be able to access normally (e.g., via swagger or TDR UI). The only thing the catalog can "expose" is data within the catalog itself.

The underlying storage system permissions are used in the catalog UI to indicate that an operation is possible, such as exporting a snapshot to a workspace. But when it comes time to do that operation, it's done using the current user's credentials, so if it turns out that they don't really have permission to do it, TDR (and the import service) won't let it happen.

pshapiro4broad, line 181: The correctness of `ForbiddenException` depends on the TDR API continuing to work the same way in the future. If it's changed to a 404 / `NotFoundException` then the code would break. So maybe it'd be better to catch the more general `ErrorReportException` here instead. Or, add an integration test that makes sure that TDR is working as this code expects it to.

common/src/main/java/bio/terra/catalog/service/DatasetService.java
```diff
@@ -170,10 +175,13 @@ public void deleteMetadata(DatasetId datasetId) {
   public String getMetadata(DatasetId datasetId) {
     var dataset = datasetDao.retrieve(datasetId);
     ensureActionPermission(dataset, SamAction.READ_ANY_METADATA);
-    return new DatasetResponse(
-            dataset, getService(dataset.storageSystem()).getDataset(dataset.storageSourceId()))
-        .convertToObject()
-        .toString();
+    StorageSystemInformation information;
+    try {
+      information = getService(dataset.storageSystem()).getDataset(dataset.storageSourceId());
```
lu-c, line 180: This code looks kind of similar - What was happening before? Would this ForbiddenException bubble its way back up to the user?

lu-c (approved):
LGTM!

common/src/main/java/bio/terra/catalog/service/DatasetService.java
pshapiro4broad: It turns out that this was the wrong exception and needed fixing. It was easy to add an integration test, so I've done that as well.

common/src/main/java/bio/terra/catalog/service/DatasetService.java
pshapiro4broad: Yes, before this change the user would see the uncaught exception in the reply. It was a wrapped API Exception so it had the HTTP status that it got from TDR, which happens to be a 403.

The exception type here was wrong in a previous comment; all `DatarepoService` APIs throw `DatarepoException`.

common/src/main/java/bio/terra/catalog/service/DatasetService.java
okotsopoulos: Okay, fair enough.  I'm convinced that this doesn't pose a permissions issue but the use of `ADMIN_INFORMATION` as the fallback variable name is still a bit confusing to me.  That could be because I'm less familiar with the repo though.

pshapiro4broad: > the use of `ADMIN_INFORMATION` as the fallback variable name is still a bit confusing to me

That's a good point, I've renamed it to `DEFAULT_INFORMATION` which should be clearer.

PR #161, opened 2023-01-03
[DC-616] Add Azure Storage Information

s-rubenstein:


common/src/main/resources/schema/development/schema.json
```diff
@@ -167,17 +167,123 @@
         }
       }
     },
-    "storageObject": {
+    "azureStorageObject": {
```
s-rubenstein, line 170: I thought that it might be cleaner to separate the enums out like this, especially since we can't have comments.

common/src/main/resources/schema/development/schema.json
```diff
@@ -167,17 +170,123 @@
...
       "type": "object",
       "properties": {
         "region": {
-          "enum": ["southamerica-west1", "us-central1", "us-east1", "us-east4", "us-west1", "us-west4", "europe-north1", "europe-west1", "europe-west4", "asia-east1", "asia-southeast1"]
+          "enum": [
+            "eastus",
```
fboulnois, line 178: Kind of a bummer that these all need to be manually specified. I wonder how often these change?

fboulnois (approved):
I like the reorganization.

common/src/main/resources/schema/development/schema.json
s-rubenstein: My guess is that they change pretty rarely, and are likely added to rather than removed from if they are. (Just because I think it is unlikely they will take down servers with people's data on them)

I agree that it is a bummer though. It is possible we could template it out and generate the json-schema at build time, but that would be a bigger change I think. (I also don't know if there are apis to get these lists automatically or not)

fboulnois: I was wondering if there was an API as well.

fboulnois: Looks like there might be? https://stackoverflow.com/questions/44143981/is-there-an-api-to-list-all-azure-regions

s-rubenstein: Neat! Why don't I add a ticket to look into schema templating to our backlog

s-rubenstein: Added: https://broadworkbench.atlassian.net/browse/DC-624

common/src/main/resources/schema/development/schema.json
```diff
@@ -167,17 +170,123 @@
...
+            "eastus2",
+            "southcentralus",
```
pshapiro4broad, line 180: It will be interesting to see how this (having a restrictive list of regions) works as the list of region changes over time. I also wonder how much value users will get from being able to do per-region filtering or searching. On the other hand it's good to have data match up across datasets and if we know the list of possible values, why not check for them.

pshapiro4broad (approved):
Looks good. Do you think we should add `storage` attributes to our integration test metadata?

I guess a more general question is, how do we verify that the schema is correct with respect to allowing valid data and disallowing invalid data as it changes over time?

We have a test schema, which is intended for testing the validation API itself, but no test data for validating the real schema.

common/src/main/resources/schema/development/schema.json
s-rubenstein: Yeah, I am most worried about data matching up between datasets. I also could see a case for filtering by region, if a research project has restrictions that data used be hosted in a particular place. (Although more often I suspect that would matter on the custodian side)

s-rubenstein:
@pshapiro4broad hmm… I wonder if we should have schema tests specifically. Let's talk at standup?

PR #162, opened 2023-01-04
DC-570, DC-622: Add example JSON for request body using openapi's `externalValue` property

pshapiro4broad:
This loads the swagger example content from an external file. It's hardcoded to work for the dataset example in for the dataset PUT operation. A future PR can improve on this work.

What works
- the code intercepts the use of `externalValue` in the PUT request body and loads an example JSON file using HTTP

What doesn't work
- can't get content type from page state
- can't get selected example from page state
- ~code to refresh page after load doesn't seem to work~
- ~example json is loaded as a string, not json object~

### DIFF
--- a/common/src/main/resources/schema/development/example.json
+++ b/common/src/main/resources/schema/development/example.json
@@ -1,0 +1,1 @@
+{"example": "json"}
--- a/common/src/main/resources/schema/development/example.json
+++ b/common/src/main/resources/schema/development/example.json
@@ -1,1 +1,197 @@
-{"example": "json"}
+{
+  "dct:title": "eiusmod",
+  "dct:description": "ex",
+  "dct:creator": "ea",
+  "dct:issued": "1972-06-02T08:42:53.0Z",
+  "TerraDCAT_ap:hasDataCollection": [
+    {
+      "dct:modified": "1957-10-06T02:22:04.0Z",
+      "dct:publisher": "qui enim dolor Excepteur Ut",
+      "dct:title": "nisi ut laboris nulla",
+      "dct:creator": "cillum voluptate anim elit dolor",
+      "dct:description": "elit reprehenderit do",
+      "dct:identifier": "eiusmod anim nulla proident Duis",
+      "dct:issued": "1979-10-26T20:48:46.0Z"
+    },
+    {
+      "dct:issued": "1988-04-08T06:33:06.0Z",
+      "dct:creator": "sed do ex mollit minim",
+      "dct:title": "veniam occaecat culpa tempor minim",
+      "dct:modified": "1943-12-25T06:59:04.0Z"
+    },
+    {
+      "dct:title": "sed nisi veniam",
+      "dct:description": "labore cillum dolor sunt et",
+      "dct:modified": "1948-03-07T11:45:21.0Z",
+      "dct:publisher": "irure",
+      "dct:issued": "1982-06-27T15:10:57.0Z",
+      "dct:identifier": "ut Ut in Duis dolore",
+      "dct:creator": "consectetur sit laboris"
+    }
+  ],
+  "dcat:accessURL": "fugiat reprehenderit eu",
+  "storage": [
+    {
+      "cloudPlatform": "gcp",
+      "qui_6a8": "magna est ipsum nostrud non",
+      "exe8_": "ullamco",
+      "dolor4bb": -93389314,
+      "mollite": 18273837.309168965,
+      "ut__": "et tempor consequat non mollit",
+      "inf6": false
+    },
+    {
+      "cloudResource": "application_deployment",
+      "region": "centralusstage",
+      "cloudPlatform": "azure"
+    },
+    {
+      "officia90": "mollit nisi fugiat",
+      "sint_886": 39193575,
+      "in708": true
+    },
+    {
+      "region": "ukwest",
+      "cloudPlatform": "azure",
+      "cloudResource": "application_deployment",
+      "dolore_25": 45212730,
+      "in_8": 21809450.76485683
+    },
+    {
+      "cloudPlatform": "azure",
+      "region": "europe",
+      "cloudResource": "synapse_workspace",
+      "proidentb0": -1669074.48199068,
+      "esseb3": true
+    }
+  ],
+  "counts": {
+    "files": -48022071,
+    "donors": 71794624,
+    "incididunt_6b": -85982146,
+    "euf6": "ut ipsum commodo magna ut",
+    "cupidatat_93": "mollit",
+    "ad0": -8528955
+  },
+  "contributors": [
+    {
+      "additionalInformation": {
+        "anim5": true,
+        "amet4b": true
+      },
+      "email": "aute fugiat amet ullamco laborum",
+      "name": "fugiat id voluptate incididunt minim"
+    },
+    {
+      "do9": 53113890.57188231,
+      "officia5df": -78638841.55110909,
+      "laboris_3d6": 71622431,
+      "Duis97": "cillum et Lorem ex"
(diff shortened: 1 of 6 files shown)


common/src/main/resources/schema/development/example.json
```diff
@@ -0,0 +1 @@
+{"example": "json"}
```
pshapiro4broad, line 1: When https://github.com/DataBiosphere/terra-data-catalog/pull/157 is merged, this can be replaced with a real example dataset.

service/src/main/resources/templates/index.html
```diff
@@ -76,31 +76,6 @@
 <script src="/webjars/swagger-ui-dist/swagger-ui-standalone-preset.js"> </script>
 <script th:inline="javascript">
 
-   // Adds support for pinning the auth bar when the user scrolls down far enough to hide the bar
```
fboulnois, line 79: Does this functionality no longer work with the custom example?

service/src/main/resources/templates/index.html
```diff
@@ -110,7 +85,48 @@
...
+    // to work with the dataset example in the PUT dataset operation.
+    const LoadExternalValuePlugin = () => {
+      return {
+        wrapComponents: {
+          // wrapping https://github.com/swagger-api/swagger-ui/blob/master/src/core/components/parameters/parameters.jsx
+          parameters: (Original, { React }) => (props) => {
+            // These hardcoded values should be replaced with code that determines the values
+            // using the user's current selection.
+            const contentType = 'application/json'
+            const activeExamplesKey = 'dataset'
```
fboulnois, line 101: Will this replace every catalog dataset with the example?

fboulnois (approved):
Two questions:

service/src/main/resources/templates/index.html
pshapiro4broad: This means that it will only affect examples (which are part of a `requestBody` parameter) that are called `dataset`. So this makes the replacement more specific.

Also note that it will only affect examples that have this path _and_ are using `externalValue`. So any examples that use `value` won't be affected.

service/src/main/resources/templates/index.html
```diff
@@ -122,12 +138,13 @@
...
-        clearValidator
+        clearValidator,
+        LoadExternalValuePlugin
       ],
       layout: 'StandaloneLayout',
-      displayOperationId: true,
-      docExpansion: 'none',
+      displayRequestDuration: true,
+      tryItOutEnabled: true,
+      defaultModelsExpandDepth: 0,
```
pshapiro4broad, line 147: These are a a few additional changes I added while working on this but are not required for this change.

- I removed `docExpansion: 'none'` because I want to have the top level tag nodes expanded by default
- `displayRequestDuration: true` shows the request duration for "try it out" requests which seems like a nice feature to have
- `tryItOutEnabled: true` changes the swagger behavior back to the old behavior, where "try it out" was enabled by default. So with this change you don't have to do an extra click before using Try it out.
- `defaultModelsExpandDepth: 0` keeps the schema models collapsed even when the tag groups are expanded.

I also removed `displayOperationId: true`. This hides the operation ID of a swagger command from the web page. So for example, the `/status` endpoint would no longer show `getStatus` next to it. The ID is really an internal thing so it seems cleaner to me to hide it, but I can add that back if others thing would be better to leave it in.

For a full list of swagger configuration options: https://swagger.io/docs/open-source-tools/swagger-ui/usage/configuration/

service/src/main/resources/templates/index.html
```diff
@@ -110,7 +85,48 @@
     };
   }
 
-  window.onload = function() {
+    // Examples cache
+    const examplesCache = {};
```
petesantos, line 89: Could this be scoped inside of `LoadExternalValuePlugin`?

service/src/main/resources/templates/index.html
```diff
@@ -110,7 +85,48 @@
...
+            const externalValue = operation.getIn(['requestBody', 'content', contentType, 'examples', activeExamplesKey, 'externalValue'])
+
+            if (externalValue) {
+              if (examplesCache[externalValue]) {
+                // If the examples cache contains the externalValue key, set the example value
+                // directly from the cache.
+                props.operation = operation.setIn(['requestBody', 'content', contentType, 'examples', activeExamplesKey, 'value'], examplesCache[externalValue])
+              } else {
+                // Fetch the external file.
+                fetch(externalValue)
```
petesantos, line 112: With fresh eyes I think I am starting to see why this isn't refreshing, but will need to play with it a little more

petesantos (approved):
LGTM. A few minor comments, feel free to take the suggestions or leave as-is

service/src/main/resources/templates/index.html
```diff
@@ -110,7 +85,48 @@
...
+                    // The code shold be able to use `oas3Actions` to refresh the UI.
+                  })
+                  .catch(e => console.error(e))
+              }
+            }
+            return React.createElement(Original, props)
+          }
+        }
+      }
+    }
```
petesantos, line 127: I put this together, and it seems to work for updating the example. I believe it is a bit more in model for react code. If you would prefer to revisit in a different PR feel free to reject the suggestion.


```suggestion
    const LoadExternalValuePlugin = () => {
      return {
        wrapComponents: {
          // wrapping https://github.com/swagger-api/swagger-ui/blob/master/src/core/components/parameters/parameters.jsx
          parameters: (Original, { React, oas3Actions }) => (props) => {
            const [example, setExample] = React.useState()
            // These hardcoded values should be replaced with code that determines the values
            // using the user's current selection.
            const {operation, pathMethod} = props
            const contentType = 'application/json'
            const activeExamplesKey = 'dataset'
            const externalValue = operation.getIn(['requestBody', 'content', contentType, 'examples', activeExamplesKey, 'externalValue'])

            // Since openAPI3 (oas3) is responsible for rendering, use a side effect to force a re-render via the oas3 UI API
            // Normally, in react the setState would be enough to get the rendered state we are after
            React.useEffect(() => {
              oas3Actions.setRequestBodyValue({ pathMethod, value: example })
            }, [example])

            if (externalValue && !example) {
                fetch(externalValue)
                  .then(res => res.text())
                  .then(data => setExample(JSON.parse(data)))
                  .catch(e => console.error(e))
            }

            // Since we need to modify the props for everything to work correctly, let's be explicit about this
            const updatedProps = example ? {
                ...props,
                ...{operation: operation.setIn(['requestBody', 'content', contentType, 'examples', activeExamplesKey, 'value'], example)}
              } : props
            return React.createElement(Original, updatedProps)
          }
        }
      }
    }

```

pshapiro4broad: Thanks, I'll give this try

pshapiro4broad: This worked to update the example, although the state still isn't quite correct (the example select menu isn't in the correct state) until the Reset button is pressed. I tried writing code to press the button automatically but couldn't get it to work.

I changed things around a little in your code, let me know what you think. I wanted to check `externalValue` separately to make it clear that the code path is only taken if `externalValue` is present. I also wanted to make the scope of `example` and `setExample` as small as possible.

service/src/main/resources/application.yml
```diff
@@ -38,7 +38,7 @@ spring:
           maxAge: 0
           mustRevalidate: true
         useLastModified: false
-      staticLocations: classpath:/api/
+      staticLocations: classpath:/api/,classpath:/schema/development/
```
pshapiro4broad, line 41: In the future, this value could be configured using the application property that specifies the active schema for a deployment. (which is currently stored in `catalog.schema.basePath`).

PR #165, opened 2023-04-27
DC-693: service startup error fix

pshapiro4broad:
This is an attempt to fix the issue where the catalog service starts up in an error state (`/status` endpoint is returning `503`) and never recovers.

One theory is that one of the status checks is taking too long, causing the status check concurrent task to no longer fire. To fix this, I've changed the concurrent task to use `scheduleWithFixedDelay()` instead of `scheduleAtFixedRate()`, which should guard against the case where a status check takes longer than the thread polling interval.

When doing this, I noticed that the timeout for the database check was using 5000 seconds. This check should be very fast so I'm changing the timeout to 1 second instead.

This change also adds unit test coverage for the database status check, and enables performing all the status checks in parallel.

### DIFF
--- a/common/src/main/java/bio/terra/catalog/service/CatalogStatusService.java
+++ b/common/src/main/java/bio/terra/catalog/service/CatalogStatusService.java
@@ -15,6 +15,8 @@ import org.springframework.stereotype.Service;
 @Service
 public class CatalogStatusService {
   private static final Logger logger = LoggerFactory.getLogger(CatalogStatusService.class);
+  /** Number of seconds to wait for a connection to the database. */
+  public static final int DB_CONNECTION_TIMEOUT = 1;
 
   private final NamedParameterJdbcTemplate jdbcTemplate;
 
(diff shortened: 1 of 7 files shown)


s-rubenstein (approved):

common/src/main/java/bio/terra/catalog/service/CatalogStatusService.java
```diff
@@ -13,31 +13,31 @@
...
   }
 
-  private SystemStatusSystems databaseStatus() {
+  @VisibleForTesting
+  SystemStatusSystems databaseStatus() {
     try {
       logger.debug("Checking database connection valid");
       return new SystemStatusSystems()
-          .ok(jdbcTemplate.getJdbcTemplate().execute((Connection conn) -> conn.isValid(5000)));
+          .ok(jdbcTemplate.getJdbcTemplate().execute((Connection conn) -> conn.isValid(1)));
```
petesantos, line 40: Thoughts on changing 1 to a constant, perhaps something named DELAY_SECONDS
Naming the number may make the intent and assumption of the developer clearer in the code for the next person.

petesantos (approved):

common/src/main/java/bio/terra/catalog/service/CatalogStatusService.java
pshapiro4broad: Sure, I can add a constant here to avoid the magic number.

PR #170, opened 2023-08-04
Test workflows

s-rubenstein:


.github/workflows/integration-tests.yml
```diff
@@ -1,16 +1,12 @@
...
-  workflow_dispatch:
+  workflow_call:
     inputs:
       environment:
-        type: choice
+        type: string
         description: 'environment to run test in'
         required: true
-        options:
-          - staging
```
pshapiro4broad, line 11: I thought this was used by the GitHub action UI to show the possible values, to see what I mean go to https://github.com/DataBiosphere/terra-data-catalog/actions/workflows/integration-tests.yml and select Run Workflow

.github/workflows/integration-tests.yml
```diff
@@ -1,16 +1,12 @@
...
-          - alpha
-          - dev
         default: 'dev'
```
snf2ye, line 10: Why are you switching to open text here?

.github/workflows/build-and-test.yml
```diff
@@ -38,18 +38,10 @@ jobs:
 
   dispatch-trivy:
     needs: [ build ]
-    runs-on: ubuntu-latest
-
+    uses: ./.github/workflows/trivy.yml
+    secrets: inherit
```
pshapiro4broad, line 42: I was reading https://docs.github.com/en/actions/using-workflows/reusing-workflows#using-inputs-and-secrets-in-a-reusable-workflow and there are ways to explicitly declare which secrets a task needs and also explicitly pass in secrets. Would that work here & make sense to do?

.github/workflows/integration-tests.yml
```diff
@@ -35,14 +31,9 @@ jobs:
...
-
-    steps:
-      - name: Fire off Trivy action
-        uses: broadinstitute/workflow-dispatch@v1
-        with:
-          workflow: Trivy
-          token: ${{ secrets.BROADBOT_TOKEN }}
+    uses: ./.github/workflows/trivy.yml
+    secrets: inherit
+    if: github.event_name == 'pull_request'
```
snf2ye, line 36: Why are you now limiting this to pull requests?

.github/workflows/integration-tests.yml
```diff
@@ -35,14 +31,9 @@ jobs:
...
     needs: [ build ]
-    runs-on: ubuntu-latest
-
-    steps:
-      - name: Fire off Trivy action
-        uses: broadinstitute/workflow-dispatch@v1
-        with:
-          workflow: Trivy
-          token: ${{ secrets.BROADBOT_TOKEN }}
+    uses: ./.github/workflows/trivy.yml
```
pshapiro4broad, line 34: I think the point of running trivy as its own workflow was so that its failure didn't fail the build, because all trivy failures we've seen so far were issues with the base image and out of our control.

If it's run this way, will it fail the build?

.github/workflows/tag.yml
```diff
@@ -1,5 +1,5 @@
 name: Tag
-on: workflow_dispatch
+on: workflow_call
```
snf2ye, line 2: @pshapiro4broad had a [good suggestion](https://github.com/DataBiosphere/java-pfb/pull/3#discussion_r1284519289) on my PR that we should define the secrets required in the action. It would probably be good to decide on a consistent pattern here!

pshapiro4broad, line 2: I think, to avoid triggering a change that causes the build to run again, you need to add the modified path (`settings.gradle`) to the `paths-ignore: [ '*.md' ]` block in `build-and-test.yml`. See https://github.com/DataBiosphere/java-pfb/blob/sh-aj-1095-publilsh/.github/workflows/build-and-test.yml#L6

pshapiro4broad (commented):
A few comments. I wish it was easier to test out changes to the GitHub workflows.

Can you check and see if other GitHub actions can be updated too? I know `broadinstitute/action-slack` was updated, not sure about other actions.

.github/workflows/integration-tests.yml
s-rubenstein: Choice seemed to not be an option for workflow_call - maybe I was misunderstanding the API docs though and there's another way to specify it?

.github/workflows/integration-tests.yml
s-rubenstein: Choice seemed to not be an option for workflow_call - maybe I was misunderstanding the API docs though and there's another way to specify it?

.github/workflows/integration-tests.yml
s-rubenstein: I think it is run as a separate stage of the workflow, so it will still run the other set of steps.

.github/workflows/integration-tests.yml
s-rubenstein: Bad copy pasta :'(

.github/workflows/integration-tests.yml
s-rubenstein: https://docs.github.com/en/enterprise-cloud@latest/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_callinputsinput_idtype

snf2ye (approved):
LGTM! Thank you!

.github/workflows/tag.yml
```diff
@@ -1,5 +1,9 @@
 name: Tag
-on: workflow_dispatch
+on:
+  workflow_call:
+    secrets:
+      token:
```
pshapiro4broad, line 5: If this is `token`, shouldn't

```
          GITHUB_TOKEN: ${{ secrets.BROADBOT_TOKEN }}
```

be

```
          GITHUB_TOKEN: ${{ secrets.token }}
```

PR #172, opened 2023-08-10
[DC-698] Try explicit auth before publishing to gcloud

s-rubenstein:
The upgrade to setup-gcloud v1 seems to encourage a separate auth step, per their docs. Trying to see if publish passes with that (need to merge to test because this is not a workflow)


snf2ye (approved):
You could test it by adding "workflow_dispatch" to the "on" argument. But, this seems like a straightforward change and it can cause things to get out of wack to run one-off publish actions. So, I'm good with just merging!

```
name: Publish and deploy
on: 
  - workflow_dispatch  
  - create
```

PR #174, opened 2024-02-28
[dc-873][dc-887] upgrade postgresql jre7

s-rubenstein:
This also updates a bunch of other plugin versions. It does not update spotbugs because that wasn't a free update, so I decided it was best to keep the changes as minimal as possible, and update that one more intentionally.

### DIFF
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -10,19 +10,19 @@ repositories {
 }
 
 dependencies {
-    implementation 'com.diffplug.spotless:spotless-plugin-gradle:6.20.0'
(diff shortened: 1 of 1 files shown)


pshapiro4broad (approved):
Looks good. I would also avoid updating spotbugs and, if anything, I would remove it as we are running sonar and it's generally be established that they are overlapping enough that we don't need to run both.

I think this also fixes the `spring-web` issue so you can tag both in the PR.

s-rubenstein:
It does! I even verified it!

snf2ye (approved):
Looks great - thank you for talking through this offline!

PR #180, opened 2024-10-30
DT-723: Increase number of retries for Integration tests

snf2ye:
https://broadworkbench.atlassian.net/browse/DT-723

### DIFF
--- a/integration/src/main/resources/configs/integration/DatasetOperations.json
+++ b/integration/src/main/resources/configs/integration/DatasetOperations.json
@@ -4,7 +4,7 @@
   "serverSpecificationFile": "catalog-local.json",
   "kubernetes": {},
   "application": {},
-  "maxRetries": 0,
+  "maxRetries": 3,
   "testScripts": [
     {
       "name": "DatasetOperations",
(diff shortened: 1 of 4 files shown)


rushtong (approved):
👍🏽

snf2ye:
I didn't see any retries in the PR run, but we can keep an eye on this for the nightly run.

rjohanek (approved):
I'm totally down to try this, but do I remember part of the conversation around catalog test failure being that its failing in setup or tear down so this may not help?

snf2ye:
> catalog test failure being that its failing in setup or tear down so this may not help?
Ah, you're totally right - that rings a bell for me. And last night's failure was indeed in the setup.
I'll pause on this change.

snf2ye:
I'm going to move forward with this change since it should not make things worse. But, I will open another ticket to look into the setup flakiness.

pshapiro4broad (commented):
I missed this PR, but the default for `maxRetries` is `3`, so an equivalent fix for this issue would be remove `maxRetries` from all configurations.

https://github.com/DataBiosphere/terra-test-runner/blob/main/src/main/java/bio/terra/testrunner/runner/config/TestConfiguration.java#L32

PR #181, opened 2024-11-01
DT-957: Reduce test flakiness - Don't use dedicated ingest service account for TDR dataset create

snf2ye:
https://broadworkbench.atlassian.net/browse/DT-957
_______
A recent staging test run failed when attempting to create a TDR dataset. The underlying issue has been noted in this TDR ticket: https://broadworkbench.atlassian.net/browse/DT-958

As a more immediate fix, we can instead use the generic TDR SA for ingest and avoid needing to create a service account. This should help with test flakiness.


s-rubenstein (approved):

rushtong (approved):
👍🏽

PR #184, opened 2025-01-09
[DT-1122] Implement `zizmor` suggestions

fboulnois:
## Addresses

https://broadworkbench.atlassian.net/browse/DT-1122

## Summary

- Do not persist credentials if not needed
- Use shell variable expansion instead of [template expansion](https://woodruffw.github.io/zizmor/audits/#template-injection)
- Include `mavenCentral` in gradle build to fix build issues


rushtong (approved):
Nice cleanup 👍🏽

rjohanek (approved):

PR #185, opened 2025-03-17
[DT-1406] Upgrade `terra-common-lib` to 1.1.38

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-1406

### Summary

Catalog doesn't depend on any advance features, so we should be able to upgrade to 1.1.38 without issue.


rushtong (approved):
👍🏽

